PR #4, opened 2020-04-22
Add work reconcile loop

qiujian16:


pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,486 @@ package controllers
...
+	controller := &ManifestWorkController{
 		manifestWorkClient:   manifestWorkClient,
-		manifestWorkInformer: manifestWorkInformer,
 		manifestWorkLister:   manifestWorkLister,
+		spokeDynamicClient:   spokeDynamicClient,
+		restMapper:           restMapper,
+		spokeResourceMonitor: map[schema.GroupVersionResource]*resourceMonitor{},
 	}
+
+	return factory.New().
```
deads2k, line 78: This one is going to need keys based on the manifest work name.  Ultimately, you may need to build something more complicated here, but this should get you going.  have a look at the hub's SpokeCluster control loop for an example.

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,486 @@ package controllers
...
+	_, manifestWorkName, err := cache.SplitMetaNamespaceKey(manifestWorkKey)
+	if err != nil {
+		return err
+	}
+
+	manifestWork, err := m.manifestWorkLister.Get(manifestWorkName)
+	resourcesOfWork := m.findResourcesOfWork(manifestWorkKey)
+
+	// Update finalizer at first
+	if manifestWork.DeletionTimestamp.IsZero() {
```
deads2k, line 101: do this before getting "resourcesOfWork"

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,486 @@ package controllers
...
 type ManifestWorkController struct {
-	manifestWorkClient   workv1client.ManifestWorkInterface
-	manifestWorkInformer workinformer.ManifestWorkInformer
-	manifestWorkLister   worklister.ManifestWorkNamespaceLister
+	manifestWorkClient workv1client.ManifestWorkInterface
+	manifestWorkLister worklister.ManifestWorkNamespaceLister
+	spokeDynamicClient dynamic.Interface
+	// spokeResourceMonitor maintains a map of informes that monitor the
+	// resources on spoke cluster
+	spokeResourceMonitor map[schema.GroupVersionResource]*resourceMonitor
```
deads2k: can I get a little more help in the doc?  I'm not clear on what the keys of this map are

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,486 @@ package controllers
...
+	// restMapper is a cached resource mapping obtained fron discovery client
+	restMapper *resource.Mapper
+}
+
+// resourceManifest is an internal type that stores info of a single manifest in one
+// reconcile loop
+type resourceManifest struct {
+	gvr         schema.GroupVersionResource
+	resourceKey string
```
deads2k: what is this?

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,486 @@ package controllers
...
+	name        string
+	namespace   string
+	manifest    *unstructured.Unstructured
+	raw         []byte
+	patch       []byte
```
deads2k: this looks odd. What's a patch for?

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,486 @@ package controllers
...
+	condition   *workapiv1.ManifestCondition
```
deads2k: this seems like an odd thing to have cached in a map.  Can you add a comment about why?

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,486 @@ package controllers
...
+	toCreate := []*resourceManifest{}
+	// toUpdate records the object to be updated on spoke
+	toUpdate := []*resourceManifest{}
+	// toDelete is a list of objectKey to be deleted on spoke cluster.
+	toDelete := []*resourceManifest{}
+
+	// The first loop iterate the spec of manifestwork and consturct manifestList
+	manifestList := []*resourceManifest{}
+	errs := []error{}
+	for i, manifestRaw := range manifestWork.Spec.Workload.Manifests {
```
deads2k: have a look to see if you can extend something like this: https://github.com/open-cluster-management/registration/pull/7/files#diff-01679b9edab3a6513f2bd7cacd5aaab1R124

It takes a set of bytes, does parsing on them inside that method, determines types, then distributes to specific types.  You'd be adding an option for generic handling as well.

Also, remember that we own that code too. If we need changes to provide better details coming out about metadata, we can add that.

deads2k: > takes a set of bytes, does parsing on them inside that method, determines types, then distributes to specific types. You'd be adding an option for generic handling as well.

If necessary, even just tolerate those specific types and we'll build out over time for a while.  We can get a lot of mileage on even *some* artifacts.

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,486 @@ package controllers
...
+		})
+
+		// Set resource meta in condition
+		resource.condition.ResourceMeta.Group = manifestResource.GroupVersionKind().Group
+		resource.condition.ResourceMeta.Version = manifestResource.GroupVersionKind().Version
+		resource.condition.ResourceMeta.Kind = manifestResource.GroupVersionKind().Kind
+		resource.condition.ResourceMeta.Name = manifestResource.GetName()
+		resource.condition.ResourceMeta.Namespace = manifestResource.GetNamespace()
+
+		// Try to generate a patch with current and modifed resource
```
deads2k: no patches.  Let's do updates.  We can try server-side-apply if we need to later.  But let's own the resources we own to start.

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,486 @@ package controllers
...
+			})
+			continue
+		}
+		if len(patch) > 0 {
+			resource.patch = patch
+			toUpdate = append(toUpdate, resource)
+		}
+	}
+
+	// 3rd loop find resources to be deleted if work does not have them in the spec any more
```
deads2k: defer this out of the immediate PR.  You will need it, but it should be computed based on prior status, don't you think?

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,486 @@ package controllers
...
+			if resource.gvr == manifest.gvr && resource.resourceKey == manifest.resourceKey {
+				delete = false
+			}
+		}
+		if delete {
+			toDelete = append(toDelete, resource)
+		}
+	}
+
+	// Finally call client to process resources.
```
deads2k: I think that apply function  I linked above can simplify a lot of this.

pkg/spoke/controllers/manifestwork_controller.go
deads2k: > defer this out of the immediate PR. You will need it, but it should be computed based on prior status, don't you think?

Yeah, definitely defer.  I'm thinking that you'll want a separate spot in status to manage this.

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,486 @@ package controllers
...
+		}
+		if !hasFinalizer {
+			manifestWork.Finalizers = append(manifestWork.Finalizers, manifestWorkFinalizer)
+			_, err := m.manifestWorkClient.Update(ctx, manifestWork, metav1.UpdateOptions{})
+			return err
+		}
+	}
+
+	// Work is deleting, we remove its related resources on spoke cluster
+	if !manifestWork.DeletionTimestamp.IsZero() {
```
deads2k, line 119: // TODO: once we make this work initially, I'd like to see the finalizer live in a different loop.  It will have different backoff considerations.

pkg/spoke/controllers/manifestwork_controller.go
deads2k: reading the code, the outcome makes sense, but I'm not seeing this as an instance variable.  it looks local to each iteration

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,486 @@ package controllers
...
-func (m *ManifestWorkController) Run(ctx context.Context, recorder events.Recorder) {
-	factory.New().
-		WithInformers(m.manifestWorkInformer.Informer()).
-		WithSync(m.sync).ResyncEvery(5*time.Minute).ToController("ManifestWorkAgent", recorder).Run(ctx, 1)
+func (m *ManifestWorkController) initMonitor(
+	gvr schema.GroupVersionResource, manifest *unstructured.Unstructured, controllerContext factory.SyncContext) *resourceMonitor {
+	monitor := m.spokeResourceMonitor[gvr]
+	if monitor == nil {
+		monitor = newMonitor(gvr, m.spokeDynamicClient, controllerContext.Queue())
+		go monitor.Run()
```
deads2k: Defer this for now.  Don't try to trigger based on change in cluster yet.  We will add it later, but not yet.

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,486 @@ package controllers
...
+		m.spokeResourceMonitor[gvr] = monitor
+	}
+	return monitor
+}
+
+func (m *ManifestWorkController) generateUpdateStatusFunc(manifests []*resourceManifest) helper.UpdateManifestWorkStatusFunc {
```
deads2k: defer out of this PR please and you can start a followup. I know it seems simple, but my head would appreciate defering what's possible

pkg/spoke/controllers/resource_monitor.go
```diff
@@ -0,0 +1,158 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/dynamic/dynamiclister"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+)
+
+type resourceMonitor struct {
```
deads2k: this looks inspired by GC.  I'm going to suggest we defer for now and consider just how dynamic we need to be.

pkg/spoke/controllers/manifestwork_controller.go
qiujian16: agree, current implementation does not cover all the cases. Need a way to store the previous state

pkg/spoke/controllers/manifestwork_controller.go
qiujian16: added

pkg/spoke/controllers/manifestwork_controller.go
qiujian16: done

pkg/helper/helper_test.go
```diff
@@ -0,0 +1,189 @@
...
+	return workapiv1.ManifestCondition{
+		ResourceMeta: workapiv1.ManifestResourceMeta{Ordinal: ordinal, Resource: resource},
+		Conditions:   conds,
+	}
+}
+
+// TestUpdateStatusCondition tests UpdateManifestWorkStatus function
+func TestUpdateStatusCondition(t *testing.T) {
+	nowish := metav1.Now()
+	beforeish := metav1.Time{Time: nowish.Add(-10 * time.Second)}
```
pmorie, line 46: :+1:

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,290 @@ package controllers
...
+		restMapper:              restMapper,
 	}
+
+	return factory.New().
+		WithInformersQueueKeyFunc(func(obj runtime.Object) string {
+			accessor, _ := meta.Accessor(obj)
+			return accessor.GetName()
+		}).
+		WithInformers(manifestWorkInformer.Informer()).
+		WithSync(controller.sync).ResyncEvery(5*time.Minute).ToController("ManifestWorkAgent", recorder)
```
pmorie, line 84: You're probably going to want to choose a higher default in time, and be able to control resync period via a CLI option; this probably works for now.

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,290 @@ package controllers
...
+	if len(errs) > 0 {
+		err = utilerrors.NewAggregate(errs)
+		klog.Errorf("Reconcile work %s fails with err %w: ", manifestWorkName, err)
+	}
+	return err
+}
+
+func (m *ManifestWorkController) cleanupResourceOfWork(work *workapiv1.ManifestWork) []error {
+	errs := []error{}
+	for _, manifest := range work.Spec.Workload.Manifests {
```
pmorie, line 176: One thing that comes to mind, making a note here but not actionable in this PR:

There will be an inherent upper limit on content in a single ManifestWork from etcd record limit, we should document that constraint in the API

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,290 @@ package controllers
...
+	}
+	mapping, err := m.restMapper.MappingForGVK(unstructuredObj.GroupVersionKind())
+	if err != nil {
+		return schema.GroupVersionResource{}, nil, fmt.Errorf("Failed to find grv from restmapping: %w", err)
+	}
+
+	return mapping.Resource, unstructuredObj, nil
+}
+
+func (m *ManifestWorkController) applyUnstructrued(data []byte, recorder events.Recorder) (*unstructured.Unstructured, bool, error) {
```
pmorie, line 261: typo: applyUnstructured

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,290 @@ package controllers
...
+		gvr, object, err := m.decodeUnstructured(manifest.Raw)
+		if err != nil {
+			klog.Errorf("Failed to decode object: %w", err)
+			errs = append(errs, err)
+			continue
+		}
+
+		err = m.spokeDynamicClient.
+			Resource(gvr).
+			Namespace(object.GetNamespace()).
```
pmorie, line 186: Is this expected to work correctly with cluster-scoped resources?

pkg/spoke/controllers/manifestwork_controller_test.go
```diff
@@ -0,0 +1,359 @@
...
+	ts *testing.T,
+	dynamicClient *fakedynamic.FakeDynamicClient,
+	workClient *fakeworkclient.Clientset,
+	kubeClient *fakekube.Clientset) {
+	workActions := workClient.Actions()
+	if len(workActions) != len(t.expectedWorkAction) {
+		ts.Errorf("Expected %d action but got %#v", len(t.expectedWorkAction), workActions)
+	}
+	for index := range workActions {
+		assertAction(ts, workActions[index], t.expectedWorkAction[index])
```
pmorie, line 257: I think we ought to add specificity to these checks.

pkg/spoke/controllers/manifestwork_controller_test.go
```diff
@@ -0,0 +1,359 @@
...
+			withExpectedWorkAction("get", "update").
+			withExpectedDynamicAction("get", "create").
+			withExpectedCondition(expectedCondition{string(workapiv1.ManifestApplied), metav1.ConditionTrue}),
+		newTestCase("multiple create&update resource").
+			withWorkManifest(newUnstructured("v1", "Secret", "ns1", "test"), newUnstructured("v1", "Secret", "ns2", "test")).
+			withSpokeObject(newSecret("test", "ns1", "value2")).
+			withExpectedWorkAction("get", "update").
+			withExpectedKubeAction("get", "delete", "create", "get", "create").
+			withExpectedCondition(expectedCondition{string(workapiv1.ManifestApplied), metav1.ConditionTrue}, expectedCondition{string(workapiv1.ManifestApplied), metav1.ConditionTrue}),
+	}
```
pmorie, line 310: I think we should have a case in this PR where the manifest isn't applied successfully.

pkg/spoke/spokeagent.go
```diff
@@ -41,14 +48,47 @@ func (o *WorkloadAgentOptions) RunWorkloadAgent(ctx context.Context, controllerC
 	// Only watch the cluster namespace on hub
 	workInformerFactory := workinformers.NewSharedInformerFactoryWithOptions(hubWorkClient, 5*time.Minute, workinformers.WithNamespace(o.SpokeClusterName))
 
+	// Build dynamic client and informer for spoke cluster
+	spokeRestConfig, err := clientcmd.BuildConfigFromFlags("", o.SpokeKubeconfigFile)
```
pmorie: Nit: in general you will save people time by commenting empty args, example: `...Flags("" /* thing */, o.SpokeKubeconfigFile)`

deads2k:
This has been simplified enough that I'm ok starting here.   I suspect we will grow complexity after, but this has potential

pkg/spoke/controllers/manifestwork_controller.go
qiujian16: yes, i think so. if namespace is "", the restclient will ignore it https://github.com/kubernetes/client-go/blob/master/dynamic/simple.go#L317-L319.

pkg/spoke/controllers/manifestwork_controller_test.go
qiujian16: done

pkg/spoke/controllers/manifestwork_controller.go
deads2k: typo still here

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,305 @@ package controllers
...
+			context.TODO(), required, metav1.CreateOptions{})
+		recorder.Eventf(fmt.Sprintf(
+			"%s Created", required.GetKind()), "Created %s/%s because it was missing", required.GetNamespace(), required.GetName())
+		return actual, true, err
+	}
+	if err != nil {
+		return nil, false, err
+	}
+
+	// Do not update unstructured type right now. Need to figure out how to merge.
```
deads2k, line 283: post merge: just do a straight update, replacing RV.  nothing fancy to start.

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,305 @@ package controllers
...
+
+	return existing, false, err
+}
+
+func (m *ManifestWorkController) generateUpdateStatusFunc(manifestStatus workapiv1.ManifestResourceStatus) helper.UpdateManifestWorkStatusFunc {
+	return func(oldStatus *workapiv1.ManifestWorkStatus) error {
+		oldStatus.ResourceStatus = manifestStatus
+		// TODO aggreagae manifest condition to generate work condition
```
deads2k, line 291: yeah, in followup

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -2,46 +2,305 @@ package controllers
...
+			return utilerrors.NewAggregate(errs)
+		}
+		return m.removeWorkFinalizer(ctx, manifestWork)
+	}
+
+	errs := []error{}
+	// Apply resources on spoke cluster.
+	resourceResults := m.applyManifest(manifestWork.Spec.Workload.Manifests, controllerContext.Recorder())
+	for index, result := range resourceResults {
+		manifestCondition := helper.FindManifestConditionByIndex(int32(index), manifestWork.Status.ResourceStatus.Manifests)
```
deads2k, line 130: minor: seems like we should write a thing that creates the conditions we want and then just call a message that does a bulk merge.  the merge logic is actually fairly complicated because it has to find any-position-to-any-position diff based on everything but ordinal as key.

deads2k:
some followups but this is a solid start.

/lgtm
/approve

pkg/spoke/controllers/manifestwork_controller.go
qiujian16: i am thinking maybe we need to check whether an update is needed, by comparing desired and existing. For instance, try generate a patch and if patch is none, do not update. Otherwise the resource might be updated in every resync.

PR #6, opened 2020-04-29
Add Dockerfile

qiujian16:


pmorie:
I caught one little thing that I will add in a follow-up, but I'm good with this and it's needed to enable promotion in the prow config.

/approve

PR #8, opened 2020-05-09
Generate work status conditions from manifest conditions

elgnay:
This is a followup of pr #4. It aggregates manifest conditions and generates work status conditions.

### DIFF
--- a/pkg/helper/helper_test.go
+++ b/pkg/helper/helper_test.go
@@ -187,3 +187,66 @@ func TestSetManifestCondition(t *testing.T) {
 		})
 	}
 }
+
+func TestMergeStatusConditions(t *testing.T) {
+	transitionTime := metav1.Now()
+
+	cases := []struct {
+		name               string
+		startingConditions []workapiv1.StatusCondition
+		newConditions      []workapiv1.StatusCondition
+		expectedConditions []workapiv1.StatusCondition
+	}{
+		{
+			name: "add status condition",
+			newConditions: []workapiv1.StatusCondition{
+				newCondition("one", "True", "my-reason", "my-message", nil),
+			},
+			expectedConditions: []workapiv1.StatusCondition{
+				newCondition("one", "True", "my-reason", "my-message", nil),
+			},
+		},
+		{
+			name: "merge status condition",
+			startingConditions: []workapiv1.StatusCondition{
+				newCondition("one", "True", "my-reason", "my-message", nil),
+			},
+			newConditions: []workapiv1.StatusCondition{
+				newCondition("one", "False", "my-reason", "my-message", nil),
+				newCondition("two", "True", "my-reason", "my-message", nil),
+			},
+			expectedConditions: []workapiv1.StatusCondition{
+				newCondition("one", "False", "my-reason", "my-message", nil),
+				newCondition("two", "True", "my-reason", "my-message", nil),
+			},
+		},
+		{
+			name: "remove useless status condition",
+			startingConditions: []workapiv1.StatusCondition{
+				newCondition("one", "False", "my-reason", "my-message", &transitionTime),
+				newCondition("two", "True", "my-reason", "my-message", nil),
+			},
+			newConditions: []workapiv1.StatusCondition{
+				newCondition("one", "False", "my-reason", "my-message", nil),
+			},
+			expectedConditions: []workapiv1.StatusCondition{
+				newCondition("one", "False", "my-reason", "my-message", &transitionTime),
+			},
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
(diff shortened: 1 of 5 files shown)


pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -295,7 +295,39 @@ func (m *ManifestWorkController) applyUnstructrued(data []byte, recorder events.
...
+			return nil
+		}
+
+		// aggregate manifest condition to generate work condition
+		applied := 0
+		for _, manifest := range manifestStatus.Manifests {
+			for _, condition := range manifest.Conditions {
+				switch {
+				case condition.Type == string(workapiv1.ManifestApplied) && condition.Status == metav1.ConditionTrue:
+					applied++
```
qiujian16: i think it needs comment here indicating how to decide the condition of manifestwork. Maybe a separate func would be better. We can reuse that func for all different condition types.  And having a counter seems not necessary. You just need a bool, and set it false if any condition status is false

pkg/spoke/controllers/manifestwork_controller_test.go
```diff
@@ -299,6 +301,26 @@ func (t *testCase) validate(
...
+
+func newManifestCondition(ordinal int32, resource string, conds ...workapiv1.StatusCondition) workapiv1.ManifestCondition {
+	return workapiv1.ManifestCondition{
+		ResourceMeta: workapiv1.ManifestResourceMeta{Ordinal: ordinal, Resource: resource},
+		Conditions:   conds,
+	}
+}
+
 // TestSync test cases when running sync
 func TestSync(t *testing.T) {
```
qiujian16: i think  you need to add the check of the  manifestwork conditions after sync in this test

pkg/helper/helper_test.go
```diff
@@ -187,3 +187,66 @@ func TestSetManifestCondition(t *testing.T) {
...
+				newCondition("one", "False", "my-reason", "my-message", nil),
+				newCondition("two", "True", "my-reason", "my-message", nil),
+			},
+			expectedConditions: []workapiv1.StatusCondition{
+				newCondition("one", "False", "my-reason", "my-message", nil),
+				newCondition("two", "True", "my-reason", "my-message", nil),
+			},
+		},
+		{
+			name: "remove useless status condition",
```
pmorie: nit: I think `remove old` would be easier to understand than `remove useless`

pkg/helper/helper_test.go
```diff
@@ -187,3 +187,66 @@ func TestSetManifestCondition(t *testing.T) {
...
+				newCondition("two", "True", "my-reason", "my-message", nil),
+			},
+			newConditions: []workapiv1.StatusCondition{
+				newCondition("one", "False", "my-reason", "my-message", nil),
+			},
+			expectedConditions: []workapiv1.StatusCondition{
+				newCondition("one", "False", "my-reason", "my-message", &transitionTime),
+			},
+		},
+	}
```
pmorie, line 236: Add a test case where the condition status doesn't change and transition time isn't updated (you could just add this to one of the existing cases too)

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -170,7 +170,7 @@ func (m *ManifestWorkController) sync(ctx context.Context, controllerContext fac
 	}
 	if len(errs) > 0 {
 		err = utilerrors.NewAggregate(errs)
-		klog.Errorf("Reconcile work %s fails with err %v: ", manifestWorkName, err)
+		klog.Errorf("Reconcile work %s fails with err: %v", manifestWorkName, err)
```
pmorie, line 173: :+1:

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -348,3 +373,21 @@ func isSameUnstructured(obj1, obj2 *unstructured.Unstructured) bool {
 
 	return equality.Semantic.DeepEqual(obj1Copy.Object, obj2Copy.Object)
 }
+
+// isAllInCondition checks status of conditions with a particular type in ManifestCondition array.
+// Return true, true only if conditions with the condition type exist and they are all in condition.
+func isAllInCondition(conditionType string, manifests []workapiv1.ManifestCondition) (inCondition bool, exists bool) {
```
pmorie: nit: `allInCondition` as a name reads a little easier to me than `isAllInCondition`

pkg/helper/helper_test.go
elgnay: I think this is already covered by test case 'remove useless status condition'.  Status of condition `one` doesn't change and its transition time is not updated after merge is done.

pmorie:
rebase needed, lgtm otherwise

PR #9, opened 2020-05-11
Update manifest condition merge logic

elgnay:
This is a followup for https://github.com/open-cluster-management/work/pull/4#discussion_r417594215.

### DIFF
--- a/pkg/helper/helpers.go
+++ b/pkg/helper/helpers.go
@@ -2,7 +2,6 @@ package helper
 
 import (
 	"context"
-	"fmt"
 	"time"
 
 	workv1client "github.com/open-cluster-management/api/client/work/clientset/versioned/typed/work/v1"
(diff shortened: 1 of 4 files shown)


pkg/helper/helpers.go
```diff
@@ -11,34 +12,126 @@ import (
...
 	}
 
-	return nil
+	return merged
 }
 
-func SetManifestCondition(conds *[]workapiv1.ManifestCondition, cond workapiv1.ManifestCondition) {
-	if conds == nil {
-		conds = &[]workapiv1.ManifestCondition{}
+func manifestResourceMetaKey(meta workapiv1.ManifestResourceMeta) string {
```
deads2k: This could actually be using a struct, right?  In fact, it could return a `ManifestResourceMeta` and just clear the ordinal, right?

pkg/helper/helpers.go
```diff
@@ -11,34 +12,126 @@ import (
...
+	for _, newCondition := range newConditions {
+		// match with ResourceMeta
+		condition, ok := metaIndex[newCondition.ResourceMeta]
+
+		// match with properties in ResourceMeta other than ordinal if not found yet
+		if !ok {
+			condition, ok = metaWithoutOridinalIndex[manifestResourceMetaKey(newCondition.ResourceMeta)]
 		}
+
+		// match with ordinal in ResourceMeta if not found yet
```
deads2k: if this happens, then I don't think you should consider them to be same.  Instead, just use the new condition.  I don't think ordinal is strong enough on its own to merit "sameness"

pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -348,3 +354,21 @@ func isSameUnstructured(obj1, obj2 *unstructured.Unstructured) bool {
 
 	return equality.Semantic.DeepEqual(obj1Copy.Object, obj2Copy.Object)
 }
+
+func buildAppliedStatusCondition(failed bool) workapiv1.StatusCondition {
```
deads2k: accept the error so you can put in the  message

deads2k:
This helps a lot, thanks.  Fairly minor comments.

PR #14, opened 2020-05-20
Fix gvk not found issue

qiujian16:
The original method may not get gvk, since kind can be empty. Change to use method in resourcehelper.


qiujian16:
close in favor of #15

PR #15, opened 2020-05-20
Fix gvk not found issue

elgnay:
This issue was introduced by https://github.com/open-cluster-management/work/pull/9

### DIFF
--- a/pkg/spoke/controllers/manifestwork_controller.go
+++ b/pkg/spoke/controllers/manifestwork_controller.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -425,3 +405,57 @@ func buildAppliedStatusCondition(err error) workapiv1.StatusCondition {
...
+
+	if object == nil {
+		return resourceMeta, err
+	}
+
+	var gvk schema.GroupVersionKind
+	var namespace, name string
+
+	switch objectType := object.(type) {
+	case *unstructured.Unstructured:
```
deads2k, line 422: this case doesn't collapse into the resourcehelper.GuessObjectGroupVersionKind(objectType) below?

elgnay: I tried and found GuessObjectGroupVersionKind() works for Unstructured object. I'll refine the code in a followup pr.

PR #16, opened 2020-05-21
Check if resource is cleaned before removing finalizer

qiujian16:
Only remove finalizer when resource is not found

relates to https://github.com/open-cluster-management/backlog/issues/2348


pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -193,6 +193,13 @@ func (m *ManifestWorkController) cleanupResourceOfWork(work *workapiv1.ManifestW
 				"Failed to delete resource %v with key %s/%s: %w",
 				gvr, object.GetNamespace(), object.GetName(), err))
 			continue
+		default:
+			// Return err since we just trigger the delete, and check if resource is not found in
```
deads2k, line 197: I'm not entirely clear on why we need to return an error when the delete was successful.  Are you worried about finalization time?

qiujian16: yes, i think so.

PR #18, opened 2020-05-21
Use same logic to get gvk of resource object

elgnay:
This is fllowup of https://github.com/open-cluster-management/work/pull/15/files#r428286111.

### DIFF
--- a/pkg/spoke/controllers/manifestwork_controller.go
+++ b/pkg/spoke/controllers/manifestwork_controller.go
@@ -411,48 +411,28 @@ func buildManifestResourceMeta(index int, object runtime.Object, restMapper *res
 		Ordinal: int32(index),
 	}
 
-	if object == nil {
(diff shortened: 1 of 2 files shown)


pkg/spoke/controllers/manifestwork_controller.go
```diff
@@ -411,48 +411,28 @@ func buildManifestResourceMeta(index int, object runtime.Object, restMapper *res
 		Ordinal: int32(index),
 	}
 
-	if object == nil {
+	if reflect.ValueOf(object).IsNil() {
```
qiujian16: I think we still need to check if object==nil before this? I think reflect.ValueOf(nil) will panic.

elgnay: Yes, you are right. Code is updated.

PR #19, opened 2020-05-21
Separate controllers

deads2k:
This separates

1. add finalizer (this was  untested before?)
2. finalize (this is straigthforward I think and I think we can use status)
3. manifestwork (this is the most complicated one)

By keeping the manifestwork controller separate, we can more safely modify it. It doesn't buy much to have them combined.  I think we should consider a separate deletion controller as well.

### DIFF
--- a/pkg/spoke/controllers/finalizercontroller/finalize_controller.go
+++ b/pkg/spoke/controllers/finalizercontroller/finalize_controller.go
@@ -1,0 +1,148 @@
+package finalizercontroller
+
+import (
+	"context"
+	"fmt"
+
+	workv1client "github.com/open-cluster-management/api/client/work/clientset/versioned/typed/work/v1"
+	workinformer "github.com/open-cluster-management/api/client/work/informers/externalversions/work/v1"
+	worklister "github.com/open-cluster-management/api/client/work/listers/work/v1"
+	workapiv1 "github.com/open-cluster-management/api/work/v1"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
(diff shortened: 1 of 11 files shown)


pkg/spoke/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -0,0 +1,148 @@
...
+	_, err := m.manifestWorkClient.Update(ctx, manifestWork, metav1.UpdateOptions{})
+	return err
+}
+
+func (m *FinalizeController) cleanupResourceOfWork(work *workapiv1.ManifestWork) []error {
+	klog.V(4).Infof("cleaning up %q", work.Name)
+
+	errs := []error{}
+
+	// TODO this can later be based on a list of all resources created by this manifest work
```
qiujian16, line 109: i think the intension is we should have another filed in manifestwork.Status to track all the resources this manifestwork have created, in addition to the resources that this manifestwork is currently managed?

Tracking resources create by the manifestwork also can be used in other scenarios, such as:
1. the name or namespace of a manifest in the manifestwork.Spec.Workload is changed.
2. a manifest is removed from manifestwork.Spec.Workload.

deads2k: > 2\. a manifest is removed from manifestwork.Spec.Workload.

Yes.  that's what I'm thinking.  An example here: https://github.com/open-cluster-management/api/pull/26

pmorie: +1.

pmorie:
Good PR, I would like @qiujian16 to lgtm when he's ready

PR #23, opened 2020-06-04
Add finalizer on role/rolebinding for work agent

elgnay:
Create controllers running on hub to add finalizer on role/rolebinding for work agent. It ensures that all manifest works in a certain cluster namespace are deleted before role/rolebinding in the same namespace is deleted.

### DIFF
--- a/pkg/hub/controllers/finalizercontroller/finalize_controller.go
+++ b/pkg/hub/controllers/finalizercontroller/finalize_controller.go
@@ -1,0 +1,245 @@
+package finalizercontroller
+
+import (
+	"context"
+	"reflect"
+	"time"
+
+	workv1client "github.com/open-cluster-management/api/client/work/clientset/versioned/typed/work/v1"
+	worklister "github.com/open-cluster-management/api/client/work/listers/work/v1"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	rbacv1informers "k8s.io/client-go/informers/rbac/v1"
+	rbacv1client "k8s.io/client-go/kubernetes/typed/rbac/v1"
+	corelisters "k8s.io/client-go/listers/core/v1"
+	rbacv1listers "k8s.io/client-go/listers/rbac/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
+)
+
+const (
+	manifestWorkFinalizer = "cluster.open-cluster-management.io/manifest-work-cleanup"
+)
+
+// RequeueDelay is time to wait before requeue the key
+var RequeueDelay = 1 * time.Minute
+
+// FinalizeController ensures all manifestworks are deleted before role/rolebinding for work
+// agent are deleted in a terminating cluster namespace.
+type FinalizeController struct {
+	roleLister         rbacv1listers.RoleLister
+	roleBindingLister  rbacv1listers.RoleBindingLister
+	rbacClient         rbacv1client.RbacV1Interface
+	namespaceLister    corelisters.NamespaceLister
+	manifestWorkClient workv1client.WorkV1Interface
+	manifestWorkLister worklister.ManifestWorkLister
+	eventRecorder      events.Recorder
+}
+
+func NewFinalizeController(
+	eventRecorder events.Recorder,
+	roleInformer rbacv1informers.RoleInformer,
+	roleBindingInformer rbacv1informers.RoleBindingInformer,
+	namespaceLister corelisters.NamespaceLister,
+	manifestWorkClient workv1client.WorkV1Interface,
+	manifestWorkLister worklister.ManifestWorkLister,
+	rbacClient rbacv1client.RbacV1Interface,
+) factory.Controller {
+
+	controller := &FinalizeController{
+		roleLister:         roleInformer.Lister(),
+		roleBindingLister:  roleBindingInformer.Lister(),
+		namespaceLister:    namespaceLister,
+		manifestWorkClient: manifestWorkClient,
+		manifestWorkLister: manifestWorkLister,
+		rbacClient:         rbacClient,
+		eventRecorder:      eventRecorder,
+	}
+
+	return factory.New().
+		WithInformersQueueKeyFunc(func(obj runtime.Object) string {
+			key, _ := cache.MetaNamespaceKeyFunc(obj)
+			return key
+		}, roleInformer.Informer(), roleBindingInformer.Informer()).
+		WithSync(controller.sync).ToController("FinalizeController", eventRecorder)
+}
+
+func (m *FinalizeController) sync(ctx context.Context, controllerContext factory.SyncContext) error {
+	key := controllerContext.QueueKey()
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		// ignore role/rolebinding whose key is not in format: namespace/name
+		return nil
+	}
+
+	ns, err := m.namespaceLister.Get(namespace)
+	if err != nil {
+		return err
+	}
+
+	role, rolebinding, err := m.getRoleAndRoleBinding(namespace, name)
+	if err != nil {
+		return err
+	}
+
+	err = m.syncRoleAndRoleBinding(ctx, controllerContext, role, rolebinding, ns)
+
+	if err != nil {
+		klog.Errorf("Reconcile role/rolebinding %s fails with err: %v", key, err)
+	}
+	return err
+}
+
+func (m *FinalizeController) syncRoleAndRoleBinding(ctx context.Context, controllerContext factory.SyncContext,
+	role *rbacv1.Role, rolebinding *rbacv1.RoleBinding, ns *corev1.Namespace) error {
+	key := controllerContext.QueueKey()
+	klog.V(4).Infof("Sync role/rolebinding %s", key)
+
+	// no work if neither role nor rolebinding has the finalizer
+	if !hasFinalizer(role, manifestWorkFinalizer) && !hasFinalizer(rolebinding, manifestWorkFinalizer) {
+		return nil
+	}
+
+	if terminated(ns) {
+		works, err := m.manifestWorkLister.ManifestWorks(ns.Name).List(labels.Everything())
+		if err != nil {
+			return err
+		}
+
+		// requeue the key if there exists any manifest work in cluster namespace
+		if len(works) != 0 {
+			controllerContext.Queue().AddAfter(key, RequeueDelay)
+			klog.V(4).Infof("Requeue role/rolebinding %q after %v", key, RequeueDelay)
+			return nil
+		}
+
+		// remove finalizer from role/rolebinding
+		if err := m.removeFinalizerFromRole(ctx, role, manifestWorkFinalizer); err != nil {
+			return err
+		}
+
+		return m.removeFinalizerFromRoleBinding(ctx, rolebinding, manifestWorkFinalizer)
+	}
+
+	if terminated(role) {
+		if err := m.removeFinalizerFromRole(ctx, role, manifestWorkFinalizer); err != nil {
+			return err
+		}
+	}
+
+	if terminated(rolebinding) {
+		if err := m.removeFinalizerFromRoleBinding(ctx, rolebinding, manifestWorkFinalizer); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (m *FinalizeController) getRoleAndRoleBinding(namespace, name string) (*rbacv1.Role, *rbacv1.RoleBinding, error) {
+	role, err := m.roleLister.Roles(namespace).Get(name)
+	if err != nil && !errors.IsNotFound(err) {
+		return nil, nil, err
+	}
+
+	rolebinding, err := m.roleBindingLister.RoleBindings(namespace).Get(name)
+	if err != nil && !errors.IsNotFound(err) {
+		return nil, nil, err
+	}
+
+	return role, rolebinding, nil
+}
+
+// removeFinalizerFromRole removes the particular finalizer from role
+func (m *FinalizeController) removeFinalizerFromRole(ctx context.Context, role *rbacv1.Role, finalizer string) error {
+	if role == nil {
+		return nil
+	}
+
+	role = role.DeepCopy()
+	if changed := removeFinalizer(role, finalizer); !changed {
+		return nil
+	}
+
+	_, err := m.rbacClient.Roles(role.Namespace).Update(ctx, role, metav1.UpdateOptions{})
+	return err
+}
+
+// removeFinalizerFromRoleBinding removes the particular finalizer from rolebinding
+func (m *FinalizeController) removeFinalizerFromRoleBinding(ctx context.Context, rolebinding *rbacv1.RoleBinding, finalizer string) error {
+	if rolebinding == nil {
+		return nil
+	}
+
+	rolebinding = rolebinding.DeepCopy()
+	if changed := removeFinalizer(rolebinding, finalizer); !changed {
+		return nil
+	}
+
+	_, err := m.rbacClient.RoleBindings(rolebinding.Namespace).Update(ctx, rolebinding, metav1.UpdateOptions{})
+	return err
+}
+
+// hasFinalizer returns true if the object has the given finalizer
+func hasFinalizer(obj runtime.Object, finalizer string) bool {
+	if obj == nil || reflect.ValueOf(obj).IsNil() {
+		return false
+	}
+
+	accessor, _ := meta.Accessor(obj)
+	for _, f := range accessor.GetFinalizers() {
+		if f == finalizer {
+			return true
+		}
+	}
+
+	return false
+}
+
+// removeFinalizer removes a finalizer from the list. It mutates its input.
+func removeFinalizer(obj runtime.Object, finalizerName string) bool {
+	if obj == nil || reflect.ValueOf(obj).IsNil() {
+		return false
+	}
+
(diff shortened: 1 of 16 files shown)


pkg/hub/controllers/finalizercontroller/add_finalizer_controller.go
```diff
@@ -0,0 +1,131 @@
...
+
+func (m *AddFinalizerController) sync(ctx context.Context, controllerContext factory.SyncContext) error {
+	key := controllerContext.QueueKey()
+
+	// ignore role/rolebinding whose name does not end with the expected suffix
+	if !strings.HasSuffix(key, keySuffix) {
+		return nil
+	}
+
+	strs := strings.Split(key, "/")
```
qiujian16: use cache.SplitMetaNamespaceKey?

pkg/hub/controllers/finalizercontroller/add_finalizer_controller.go
```diff
@@ -0,0 +1,131 @@
...
+	controller := &AddFinalizerController{
+		gvr:           gvr,
+		lister:        informer.Lister(),
+		clusterLister: clusterLister,
+		dynamicClient: dynamicClient,
+	}
+
+	return factory.New().
+		WithInformersQueueKeyFunc(func(obj runtime.Object) string {
+			accessor, _ := meta.Accessor(obj)
```
qiujian16: use cache.MetaNamespaceKeyFunc?

pkg/hub/controllers/finalizercontroller/add_finalizer_controller.go
```diff
@@ -0,0 +1,131 @@
...
+	// ignore role/rolebinding whose name does not end with the expected suffix
+	if !strings.HasSuffix(key, keySuffix) {
+		return nil
+	}
+
+	strs := strings.Split(key, "/")
+	if len(strs) != 2 {
+		return nil
+	}
+	namespace, name, clusterName := strs[0], strs[1], strs[1][:strings.LastIndex(strs[1], ":")]
```
qiujian16: I think you can use TrimPrefix  and TrimSuffix here? And need a comments

pkg/hub/controllers/finalizercontroller/add_finalizer_controller.go
```diff
@@ -0,0 +1,131 @@
...
+	obj, err := m.lister.ByNamespace(namespace).Get(name)
+	if errors.IsNotFound(err) {
+		// resource not found, could have been deleted, do nothing.
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	u, ok := obj.(*unstructured.Unstructured)
```
qiujian16: Why unstructured? i think it should be role and rolebinding?

pkg/hub/manager.go
```diff
@@ -0,0 +1,78 @@
...
+	roleBindingGVR := schema.GroupVersionResource{
+		Group:    "rbac.authorization.k8s.io",
+		Version:  "v1",
+		Resource: "rolebindings",
+	}
+	roleBindingInformer := dynamicInformers.ForResource(roleBindingGVR)
+	addFinalizerControllerForRoleBinding := finalizercontroller.NewAddFinalizerController(controllerContext.EventRecorder, roleBindingGVR,
+		roleBindingInformer, clusterLister, dynamicClient, "AddFinalizerControllerForRoleBinding")
+
+	finalizeControllerForRoleBinding := finalizercontroller.NewFinalizeController(controllerContext.EventRecorder, roleBindingGVR,
```
qiujian16: i see the motivation to use dynamicClient...But I think we could have one addFinalizerController to update both finalizer of role/rolebinding, and another finalizerController to remove finalizers.

qiujian16:
needs an integration test on this.

pkg/hub/controllers/finalizercontroller/add_finalizer_controller.go
```diff
@@ -0,0 +1,171 @@
...
+// finalizer on role/rolebinding for work agent in spoke cluster namespace.
+type AddFinalizerController struct {
+	roleLister        rbacv1listers.RoleLister
+	roleBindingLister rbacv1listers.RoleBindingLister
+	rbacClient        rbacv1client.RbacV1Interface
+	clusterLister     clusterv1listers.SpokeClusterLister
+}
+
+// NewAddFinalizerController returns a AddFinalizerController
+func NewAddFinalizerController(
```
deads2k: fwiw, I think that simply adding it to your resource is enough.  if a user changes it, they shot themselves in the foot, but I don't they expand their impact/control

pkg/hub/controllers/finalizercontroller/add_finalizer_controller.go
```diff
@@ -0,0 +1,171 @@
...
+	if len(errs) > 0 {
+		err = utilerrors.NewAggregate(errs)
+		klog.Errorf("Reconcile role/rolebinding %s fails with err: %v", key, err)
+		return err
+	}
+	return nil
+}
+
+// syncRole adds finalizer on role for work agent in the spoke cluster namespace
+func (m *AddFinalizerController) syncRole(ctx context.Context, role *rbacv1.Role) error {
```
deads2k: `addRoleFinalizer` I'd think

pkg/hub/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -0,0 +1,249 @@
...
+
+	return factory.New().
+		WithInformersQueueKeyFunc(func(obj runtime.Object) string {
+			key, _ := cache.MetaNamespaceKeyFunc(obj)
+			return key
+		}, roleInformer.Informer(), roleBindingInformer.Informer()).
+		WithSync(controller.sync).ToController("FinalizeController", recorder)
+}
+
+func (m *FinalizeController) sync(ctx context.Context, controllerContext factory.SyncContext) error {
```
deads2k, line 74: I see two primary cases.
1. client deletes role or rolebinding, but does not delete the namespace.  In this case, allow them to be deleted.  The controller should recreate it.
2. client deletes the namespace, which causes the roles and rolebindings to be deleted.  In this case, keep the role and rolebinding until the manifestworks are gone.

If we do it like that, I think the logic here needs some updates so that a delete of replaceable roles and rolebindings does not result in deletes of manifestworks.

I think you should also consider how orphaning a cluster would work.  I suspect it involves first removing the spoke controllers, but it's a think to consider.

qiujian16: yes, I did not consider that client delete role/rolebinding. I think we can check if all manifestworks  are all deleted then remove finalizer, instead of deleting manifestwork explicitly. Namespace deletion will trigger manifestwork deletion. Does it make sense?

pkg/hub/controllers/finalizercontroller/add_finalizer_controller.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	// don't add finalizer to instances that already have it
+	for i := range role.Finalizers {
+		if role.Finalizers[i] == manifestWorkFinalizer {
+			return nil
+		}
+	}
+
+	// if this conflicts, we'll simply try again later
```
pmorie: nit: it's best to avoid using words like 'simply', they are subjective. if you just drop 'simply' the comment will be informative.

pkg/hub/controllers/finalizercontroller/finalize_controller.go
pmorie: > I see two primary cases.

I'm +1 on what David articulated here and on outlining how to orphan a cluster.

pkg/hub/controllers/finalizercontroller/add_finalizer_controller.go
qiujian16: the role/rolebinding is currently applied by registration controller. Do you think it is better we should apply work's role/rolebindnig in work controller?

elgnay: The suggestion makes sense. Now the resources (role/rolebinding) are created by controller in registration repo. That means the finalizer will be added by controller in registration repo and removed by controller in work repo. This may not be a good idea. How about we still use AddFinalizerController to add finalizer to role/rolebinding until we move the creation of role/rolebinding from registration repo to work repo?

pkg/hub/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -0,0 +1,278 @@
...
+func (m *FinalizeController) syncRoleAndRoleBinding(ctx context.Context, controllerContext factory.SyncContext,
+	role *rbacv1.Role, rolebinding *rbacv1.RoleBinding, ns *corev1.Namespace, listWorksFunc listManifestWorksFunc) error {
+	klog.V(4).Infof("Sync role/rolebinding %s/%s%s", ns.Name, ns.Name, keySuffix)
+	// no work if neither role nor rolebinding exists
+	if role == nil && rolebinding == nil {
+		return nil
+	}
+
+	// no work if no finalizer found on either role or rolebinding
+	if !hasFinalizer(role, rolebinding) {
```
deads2k: just because a role doesn't have the finalizer, doesn't mean you shouldn't deal with the rolebinding.  Handle these separately.

pkg/hub/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -0,0 +1,278 @@
...
+		return nil
+	}
+
+	if ns.DeletionTimestamp == nil || ns.DeletionTimestamp.IsZero() {
+		// ignore role if it is not terminating
+		if role != nil && (role.DeletionTimestamp == nil || role.DeletionTimestamp.IsZero()) {
+			role = nil
+		}
+
+		// ignore rolebinding if it is not terminating
```
deads2k: ignoring isn't right.  You need to remove the finalizer so that the rolebinding will be recreated.

pkg/hub/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -0,0 +1,278 @@
...
+	for _, work := range works {
+		if len(work.Finalizers) == 0 {
+			continue
+		}
+
+		if work.DeletionTimestamp == nil || work.DeletionTimestamp.IsZero() {
+			continue
+		}
+
+		if now.Sub(work.DeletionTimestamp.Time) < DeletionTimeout {
```
deads2k: we don't normally place timeouts on finalziers

pkg/hub/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+	}
+
+	if len(errs) > 0 {
+		return utilerrors.NewAggregate(errs)
+	}
+	return nil
+}
+
+func (m *FinalizeController) cleanUpManifestWorks(listWorksFunc listManifestWorksFunc) (bool, error) {
```
deads2k: doc what this is doing for me.  I expected to simply count the manifests works in the namespace and if there was more than zero, don't remove the finalizer on the role or rolebinding.

pkg/hub/controllers/finalizercontroller/finalize_controller.go
elgnay: The comment here is incorrect. It should skip if neither role nor rolebinding has the finalizer

pkg/hub/controllers/finalizercontroller/finalize_controller.go
elgnay: In a non-terminating namespace, I think, we should do nothing on a role/rolebinding if it is not being deleted by client. Instead, if it is being deleted (DeletionTimestamp is set), the finalizer should be removed.

pkg/hub/controllers/finalizercontroller/finalize_controller.go
elgnay: This func is removed along with the timeout feature.

pkg/hub/controllers/finalizercontroller/add_finalizer_controller.go
deads2k: > The suggestion makes sense. Now the resources (role/rolebinding) are created by controller in registration repo. That means the finalizer will be added by controller in registration repo and removed by controller in work repo. This may not be a good idea. How about we still use AddFinalizerController to add finalizer to role/rolebinding until we move the creation of role/rolebinding from registration repo to work repo

The matching logic here isn't significantly less fragile than simply adding the finalizer to the manifests in another repo.  Now, given usage, you may be better off putting this in the registration repo and simply listing the manifestworks until there are none.

pkg/hub/controllers/finalizercontroller/add_finalizer_controller.go
```diff
@@ -0,0 +1,158 @@
...
+	if !strings.HasSuffix(key, keySuffix) {
+		return nil
+	}
+
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		// ignore role/rolebinding whose key is not in format: namespace/name
+		return nil
+	}
+	clusterName := strings.TrimSuffix(name, keySuffix)
```
deads2k: logic like this concerns me and it means that the linkage isn't really any better since the two repos are still coordinating

pkg/hub/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -0,0 +1,230 @@
...
+	rbacv1informers "k8s.io/client-go/informers/rbac/v1"
+	rbacv1client "k8s.io/client-go/kubernetes/typed/rbac/v1"
+	corelisters "k8s.io/client-go/listers/core/v1"
+	rbacv1listers "k8s.io/client-go/listers/rbac/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
+)
+
+// RequeueDelay is time to wait before requeue the key
+var RequeueDelay = 2 * time.Second
```
deads2k: This is too tight.  There is a method called `c.queue.AddRateLimited(key)`  I think that's what you want.

pkg/hub/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -0,0 +1,230 @@
...
+		if role != nil && (role.DeletionTimestamp == nil || role.DeletionTimestamp.IsZero()) {
+			role = nil
+		}
+
+		// skip the rolebinding if it is in a non-terminating namespace and not being deleted
+		if rolebinding != nil && (rolebinding.DeletionTimestamp == nil || rolebinding.DeletionTimestamp.IsZero()) {
+			rolebinding = nil
+		}
+	} else {
+		works, err := listWorksFunc()
```
deads2k: just inline this `m.manifestWorkLister.ManifestWorks(ns.Name).List(labels.Everything())`

pkg/hub/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -0,0 +1,230 @@
...
+
+	rolebinding, err := m.roleBindingLister.RoleBindings(namespace).Get(name)
+	if err != nil && !errors.IsNotFound(err) {
+		return nil, nil, err
+	}
+
+	return role, rolebinding, nil
+}
+
+func hasFinalizer(role *rbacv1.Role, rolebinding *rbacv1.RoleBinding) bool {
```
deads2k: update this to be a function on ObjectMeta and then call it twice please.  I have trouble reading the code.

pkg/hub/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -0,0 +1,230 @@
...
+	// ignore role/rolebinding if namespace does not match with cluster name
+	if namespace != clusterName {
+		return nil
+	}
+
+	ns, err := m.namespaceLister.Get(namespace)
+	if err != nil {
+		return err
+	}
+
```
deads2k, line 86: proposal

```go
if !ns.terminating{
    // unconditionally remove the finalizer
    return
}

numManifestWorksRemaining :=  m.manifestWorkLister.ManifestWorks(ns.Name).List(labels.Everything())
if numManifestWorksRemaining > 0{
    queue.AddRateLimited(key)
    return nil
}

if objectMetaHasFinalizer(role){
    remove finalizer
}

if objectMetaHasFinalizer(rolebinding){
    remove finalizer
}
```

This has minimal nesting and early returns which made the code very easy to reason about.

pkg/hub/controllers/finalizercontroller/add_finalizer_controller.go
elgnay: I'll create a PR on registration repo to make the change. But I'm still wondering if we have any use case which involves registration repo only. In that case, we may have trouble in deleting the role/rolebinding for work agent due to the finalizer.

pkg/hub/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -0,0 +1,245 @@
...
+	if terminated(rolebinding) {
+		if err := m.removeFinalizerFromRoleBinding(ctx, rolebinding, manifestWorkFinalizer); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (m *FinalizeController) getRoleAndRoleBinding(namespace, name string) (*rbacv1.Role, *rbacv1.RoleBinding, error) {
+	role, err := m.roleLister.Roles(namespace).Get(name)
```
qiujian16, line 146: i'm thinking whether we should filter the role by rules, so we only add finalizer to the roles who have rules to update/updatestatus of manifestworks?

pkg/hub/controllers/finalizercontroller/finalize_controller_test.go
```diff
@@ -0,0 +1,187 @@
...
+					t.Fatal(spew.Sdump(actions))
+				}
+			},
+		},
+		{
+			name:        "remove finalizer from role/rolebinding within terminating namespace",
+			role:        hubtesting.NewRole("cluster1", "cluster1:spoke-work", []string{manifestWorkFinalizer}, true),
+			roleBinding: hubtesting.NewRoleBinding("cluster1", "cluster1:spoke-work", []string{manifestWorkFinalizer}, true),
+			namespace:   hubtesting.NewNamespace("cluster1", true),
+			validateRabcActions: func(t *testing.T, actions []clienttesting.Action) {
```
pmorie, line 72: typo: rbac

qiujian16:
/hold
reconsider move this into registration controller

pmorie:
Closing since registration/48 handles this

PR #25, opened 2020-06-16
Support foreground deletion

elgnay:
To fix issue https://github.com/open-cluster-management/backlog/issues/2348, support foreground deletion


pkg/spoke/controllers/deletioncontroller/deletion_controller.go
```diff
@@ -117,36 +134,68 @@ func (m *StaleManifestDeletionController) syncManifestWork(ctx context.Context,
...
+			errs = append(errs, fmt.Errorf(
+				"Failed to get resource %v with key %s/%s: %w",
+				gvr, resource.Namespace, resource.Name, err))
+			deletingResources = append(deletingResources, resource)
 			continue
 		}
-		klog.V(2).Infof("Successfully delete resource %v with key %s/%s", gvr, resource.Namespace, resource.Name)
+
+		// delete the resource which exists and is not being deleted
+		if deletionTimestamp := u.GetDeletionTimestamp(); deletionTimestamp == nil || deletionTimestamp.IsZero() {
```
deads2k: invert this check and `continue` to reduce nesting.

pkg/spoke/controllers/deletioncontroller/deletion_controller.go
```diff
@@ -117,36 +134,68 @@ func (m *StaleManifestDeletionController) syncManifestWork(ctx context.Context,
...
+
+	// requeue the work if there exists deleting applied resources
+	if len(deletingResources) != 0 {
+		requeueWork()
+	}
+
+	return nil
 }
 
 // deleteAppliedResources deletes all applied resources in the given resource meta array.
```
deads2k: godoc should tell me what's being returned.

pkg/spoke/controllers/deletioncontroller/deletion_controller.go
```diff
@@ -23,6 +24,9 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 )
 
+// RequeueDelay is the time to wait before requeuing the work
+var RequeueDelay = 1 * time.Second
```
deads2k: too tight.  If it's not backing off, check every minute

pkg/spoke/controllers/deletioncontroller/deletion_controller.go
```diff
@@ -117,36 +134,68 @@ func (m *StaleManifestDeletionController) syncManifestWork(ctx context.Context,
...
 	// for retrying update without reassessing the status can cause overwriting of valid information.
 	manifestWork.Status.AppliedResources = appliedResources
 	_, err := m.manifestWorkClient.UpdateStatus(ctx, manifestWork, metav1.UpdateOptions{})
-	return err
+	if err != nil {
+		return err
+	}
+
+	// requeue the work if there exists deleting applied resources
+	if len(deletingResources) != 0 {
```
deads2k: Doing it this way prevents orphaning of resources and prevents adoption by other manifest work without an outage.  This can result in perma-stuck resources.  Be sure you mean this.

pkg/spoke/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -20,6 +21,9 @@ import (
 	"k8s.io/klog"
 )
 
+// RequeueDelay is the time to wait before requeuing the work
+var RequeueDelay = 1 * time.Second
```
deads2k: too tight. one minute

pkg/spoke/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -117,31 +132,50 @@ func (m *FinalizeController) syncManifestWork(ctx context.Context, originalManif
...
-				"Failed to delete resource %v with key %s/%s: %w",
+				"Failed to get resource %v with key %s/%s: %w",
 				gvr, appliedResource.Namespace, appliedResource.Name, err))
+			appliedResources = append(appliedResources, appliedResource)
 			continue
 		}
-		klog.Infof("Successfully delete resource %v with key %s/%s", gvr, appliedResource.Namespace, appliedResource.Name)
+
+		// delete the resource which exists and is not being deleted
+		if deletionTimestamp := u.GetDeletionTimestamp(); deletionTimestamp == nil || deletionTimestamp.IsZero() {
```
deads2k: invert check and continue to reduce nesting

pkg/spoke/controllers/deletioncontroller/deletion_controller.go
elgnay: To fix this issue, I suggest adding a new optional property, like `DeletionTimestamp`, into AppliedManifestResourceMeta struct. The property will be set once controller calls the delete api for the applied resource successfully. StaleManifestDeletionController will only delete applied resource of stale manifest whose `DeletionTimestamp` is not set, which ensures deletion of applied resource is done only once by controller. When checking the existence of deleting resource, if the controller finds the `DeletionTimestamp` of this resource is not set, it will treat the original resource deleted and then remove it from `ManifestWork.Status.AppliedResources`.

pkg/spoke/controllers/finalizercontroller/add_finalizer_controller_test.go
```diff
@@ -13,7 +13,7 @@ import (
 	clienttesting "k8s.io/client-go/testing"
 )
 
-func TestAddFinalizer(t *testing.T) {
+func PTestAddFinalizer(t *testing.T) {
```
pmorie: Was this intentional?

pkg/spoke/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -117,31 +132,50 @@ func (m *FinalizeController) syncManifestWork(ctx context.Context, originalManif
...
+				Namespace(appliedResource.Namespace).
+				Delete(context.TODO(), appliedResource.Name, metav1.DeleteOptions{})
+			if errors.IsNotFound(err) {
+				continue
+			}
+			if err != nil {
+				errs = append(errs, fmt.Errorf(
+					"Failed to delete resource %v with key %s/%s: %w",
+					gvr, appliedResource.Namespace, appliedResource.Name, err))
+			}
```
pmorie: do we want to `continue` here as well? as it stands, this will still be added to `appliedResources`

pkg/spoke/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -117,31 +132,50 @@ func (m *FinalizeController) syncManifestWork(ctx context.Context, originalManif
 func (m *FinalizeController) cleanupResourceOfWork(work *workapiv1.ManifestWork) ([]workapiv1.AppliedManifestResourceMeta, []error) {
```
pmorie: I think it's time for some godoc to say what this method does

pkg/spoke/controllers/finalizercontroller/finalize_controller.go
pmorie: Nevermind, after re-reading it's more clear what this is doing. Let's make sure this info is encoded in the godoc so we know what `appliedResources` should contain as a return value.

pkg/spoke/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -61,12 +67,15 @@ func (m *FinalizeController) sync(ctx context.Context, controllerContext factory
...
+	return m.syncManifestWork(ctx, manifestWork, func() {
+		controllerContext.Queue().AddAfter(manifestWorkName, RequeueDelay)
+	})
 }
 
-// syncManifestWork ensures that when a manifestwork has been deleted, everything it created is also deleted before removing
-// the finalizer
-func (m *FinalizeController) syncManifestWork(ctx context.Context, originalManifestWork *workapiv1.ManifestWork) error {
+// syncManifestWork ensures that when a manifestwork has been deleted, everything it created is also deleted.
+// Foreground deletion is implemented, which means all ressources created will be deleted and become invisiable
```
pmorie: typo: resources

Also, this isn't exactly accurate:

> and become invisible in the k8s cluster

Deleted but unfinalized stuff is still visible in the control plane - it is only fully gone once it's finalized. So, after this resources will be deleted but may be unfinalized for arbitrary periods.

qiujian16:
/hold
until we figure out whether criticalresource works as expected.

deads2k:
Let's get this PR back up and running.  I think we can try this as default behavior.  I suspect we may want a field in spec controlling it

/hold cancel

pkg/spoke/controllers/deletioncontroller/deletion_controller.go
```diff
@@ -112,41 +134,98 @@ func (m *StaleManifestDeletionController) syncManifestWork(ctx context.Context,
...
+			deletingResources = append(deletingResources, resource)
+			continue
+		}
+
 		switch {
-		case errors.IsNotFound(err):
-			// no-oop
-		case err != nil:
-			errs = append(errs, fmt.Errorf("Failed to delete resource %v with key %s/%s: %w", gvr, resource.Namespace, resource.Name, err))
+		case resource.UID != "" && resource.UID == string(u.GetUID()):
```
qiujian16: this seems weird. Is this condition happens when deletion failed in the previous loop? It seems to me if such condition happens, we did not do anything and all following loops falls into this condition?

elgnay: This case should never happen if everything goes well. I've removed it from code.

pkg/spoke/controllers/deletioncontroller/deletion_controller.go
```diff
@@ -66,10 +73,16 @@ func (m *StaleManifestDeletionController) sync(ctx context.Context, controllerCo
...
+		controllerContext.Queue().AddAfter(manifestWorkName, m.rateLimiter.When(manifestWorkName))
+	}, func() {
+		m.rateLimiter.Forget(manifestWorkName)
+	})
 }
 
-func (m *StaleManifestDeletionController) syncManifestWork(ctx context.Context, originalManifestWork *workapiv1.ManifestWork) error {
+// syncManifestWork collects latest applied resources from ManifestWork.Status.ResourceStatus and merges them with
+// existing applied resources. It also deletes applied resources for stale manifests
+func (m *StaleManifestDeletionController) syncManifestWork(ctx context.Context, originalManifestWork *workapiv1.ManifestWork, requeueWork requeueWorkFunc, forgetWork forgetWorkFunc) error {
```
deads2k: passing in requeueWork and forgetWork seems unnecessary since it could exist on `m`.  I'm guessing you've set this up for unit testing, so I don't object to the idea of stubs, but I would suggest something like

```go
type StaleManifestDeletionController struct{
  // stuff
  requeueWorkUnitTestFn func()
  forgetWorkUnitTestFn func()
}

func (m *Stale) requeueWork(){
  if m.requeueWorkUnitTestFn != nil{
    // unit testing
    m.requeueWorkUnitTestFn()
    return
  }

  controllerContext.Queue().AddAfter(manifestWorkName, m.rateLimiter.When(manifestWorkName))
}

```

pkg/spoke/controllers/deletioncontroller/deletion_controller.go
```diff
@@ -96,6 +109,15 @@ func (m *StaleManifestDeletionController) syncManifestWork(ctx context.Context,
 		})
 	}
 
+	// delete applied resources which are no longer maintained by manifest work
+	noLongerMaintainedResources := findUntrackedResources(manifestWork.Status.AppliedResources, appliedResources)
+	deletingResources, errs := m.deleteAppliedResources(noLongerMaintainedResources)
```
deads2k: s/deletingResources/resourcesPendingFinalization/g

pkg/spoke/controllers/deletioncontroller/deletion_controller.go
```diff
@@ -112,41 +134,95 @@ func (m *StaleManifestDeletionController) syncManifestWork(ctx context.Context,
 		}
 	})
 
-	// no work to do if applied resources are not changed
 	if reflect.DeepEqual(manifestWork.Status.AppliedResources, appliedResources) {
+		// requeue the work if there exists deleting applied resources
+		if len(deletingResources) != 0 {
```
deads2k: this doesn't need to be inside this predicate, allowing an if/else above instead of the logic being 10 lines down from here.

pkg/spoke/controllers/deletioncontroller/deletion_controller.go
```diff
@@ -112,41 +134,95 @@ func (m *StaleManifestDeletionController) syncManifestWork(ctx context.Context,
...
+				"Failed to get resource %v with key %s/%s: %w",
+				gvr, resource.Namespace, resource.Name, err))
+			deletingResources = append(deletingResources, resource)
+			continue
+		}
+
+		// leave deleting resource untouched
+		if deletionTimestamp := u.GetDeletionTimestamp(); deletionTimestamp != nil && !deletionTimestamp.IsZero() {
+			// set the UID even if the resource is deleted by other client
+			if resource.UID == "" {
```
deads2k: kube standard is `len(resource.UID) == 0`

pkg/spoke/controllers/deletioncontroller/deletion_controller.go
```diff
@@ -112,41 +134,95 @@ func (m *StaleManifestDeletionController) syncManifestWork(ctx context.Context,
...
+				resource.UID = string(u.GetUID())
+			}
+			deletingResources = append(deletingResources, resource)
+			continue
+		}
+
+		if resource.UID != "" {
```
deads2k: Also, please do an equality test ``len(resource.UID) != 0 && resource.UID != u.UID` instead.  And mark the top of the block with "// forget this item if the UID does not match the resource UID we previously deleted"

pkg/spoke/controllers/deletioncontroller/deletion_controller.go
```diff
@@ -112,41 +134,95 @@ func (m *StaleManifestDeletionController) syncManifestWork(ctx context.Context,
...
+			// have already been deleted and then re-created by other client. Manifest work should not track it
+			// any longer
+			continue
+		}
+
+		// delete the resource which exists and is not being deleted
+		err = m.spokeDynamicClient.
+			Resource(gvr).
+			Namespace(resource.Namespace).
+			Delete(context.TODO(), resource.Name, metav1.DeleteOptions{})
```
deads2k: add a UID precondition here and check the appropriate error if the precondition fails.

pkg/spoke/controllers/deletioncontroller/deletion_controller.go
```diff
@@ -112,41 +134,95 @@ func (m *StaleManifestDeletionController) syncManifestWork(ctx context.Context,
...
+		}
+		if err != nil {
+			errs = append(errs, fmt.Errorf(
+				"Failed to delete resource %v with key %s/%s: %w",
+				gvr, resource.Namespace, resource.Name, err))
+			continue
+		}
+
+		// set UID of applied resource once deletion is successful
+		resource.UID = string(u.GetUID())
```
deads2k: make a copy before mutation please.

pkg/spoke/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -61,16 +68,21 @@ func (m *FinalizeController) sync(ctx context.Context, controllerContext factory
...
+	})
 }
 
-// syncManifestWork ensures that when a manifestwork has been deleted, everything it created is also deleted before removing
-// the finalizer
-func (m *FinalizeController) syncManifestWork(ctx context.Context, originalManifestWork *workapiv1.ManifestWork) error {
+// syncManifestWork ensures that when a manifestwork has been deleted, everything it created is also deleted.
+// Foreground deletion is implemented, which means all resources created will be deleted and finalized
+// before removing finalizer from manifestwork
+func (m *FinalizeController) syncManifestWork(ctx context.Context, originalManifestWork *workapiv1.ManifestWork, requeueWork requeueWorkFunc, forgetWork forgetWorkFunc) error {
```
deads2k: same comment as above about accepting functions

pkg/spoke/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -89,59 +101,110 @@ func (m *FinalizeController) syncManifestWork(ctx context.Context, originalManif
 	var err error
 
 	// Work is deleting, we remove its related resources on spoke cluster
-	remaining, errs := m.cleanupResourceOfWork(manifestWork)
-	if len(manifestWork.Status.AppliedResources) != len(remaining) {
+	deletingResources, errs := m.cleanupResourceOfWork(manifestWork)
+	updated := false
```
deads2k: `updatedManifestWork`

pkg/spoke/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -89,59 +101,110 @@ func (m *FinalizeController) syncManifestWork(ctx context.Context, originalManif
...
+	forgetWork()
 
 	removeFinalizer(manifestWork, manifestWorkFinalizer)
 	_, err = m.manifestWorkClient.Update(ctx, manifestWork, metav1.UpdateOptions{})
 	return err
 }
 
+// cleanupResourceOfWork deletes applied resources of the manifest work and returns those
+// which are deleted but not finalized yet
 func (m *FinalizeController) cleanupResourceOfWork(work *workapiv1.ManifestWork) ([]workapiv1.AppliedManifestResourceMeta, []error) {
```
deads2k: is this function essentially the same as above?

pkg/spoke/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -89,59 +101,110 @@ func (m *FinalizeController) syncManifestWork(ctx context.Context, originalManif
...
+				"Failed to get resource %v with key %s/%s: %w",
+				gvr, appliedResource.Namespace, appliedResource.Name, err))
+			appliedResources = append(appliedResources, appliedResource)
+			continue
+		}
+
+		// leave deleting resource untouched
+		if deletionTimestamp := u.GetDeletionTimestamp(); deletionTimestamp != nil && !deletionTimestamp.IsZero() {
+			// set the UID even if the resource is not deleted by controller
+			if appliedResource.UID == "" {
```
deads2k: above comment applies

pkg/spoke/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -89,59 +101,110 @@ func (m *FinalizeController) syncManifestWork(ctx context.Context, originalManif
...
+				appliedResource.UID = string(u.GetUID())
+			}
+			appliedResources = append(appliedResources, appliedResource)
+			continue
+		}
+
+		if appliedResource.UID != "" {
```
deads2k: above comment applies

pkg/spoke/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -89,59 +101,110 @@ func (m *FinalizeController) syncManifestWork(ctx context.Context, originalManif
...
+		if errors.IsNotFound(err) {
+			continue
+		}
+		if err != nil {
 			errs = append(errs, fmt.Errorf(
 				"Failed to delete resource %v with key %s/%s: %w",
 				gvr, appliedResource.Namespace, appliedResource.Name, err))
 			continue
 		}
-		klog.Infof("Successfully delete resource %v with key %s/%s", gvr, appliedResource.Namespace, appliedResource.Name)
```
deads2k, line 141: your call, but you don't want a V(2) or a V(4)?

pkg/spoke/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -89,59 +101,110 @@ func (m *FinalizeController) syncManifestWork(ctx context.Context, originalManif
...
+
+		// set UID of applied resource once deletion is successful
+		appliedResource.UID = string(u.GetUID())
```
deads2k: same comment about mutation

pkg/spoke/controllers/finalizercontroller/finalize_controller.go
```diff
@@ -155,3 +218,8 @@ func removeFinalizer(manifestWork *workapiv1.ManifestWork, finalizerName string)
 	}
 	manifestWork.Finalizers = newFinalizers
 }
+
+func newDeletionTimestamp() *metav1.Time {
```
deads2k: cruft?

pkg/spoke/controllers/deletioncontroller/deletion_controller.go
```diff
@@ -112,41 +134,95 @@ func (m *StaleManifestDeletionController) syncManifestWork(ctx context.Context,
...
+			continue
+		}
+
+		// leave deleting resource untouched
+		if deletionTimestamp := u.GetDeletionTimestamp(); deletionTimestamp != nil && !deletionTimestamp.IsZero() {
+			// set the UID even if the resource is deleted by other client
+			if resource.UID == "" {
+				resource.UID = string(u.GetUID())
+			}
+			deletingResources = append(deletingResources, resource)
```
deads2k: check to see if you're deleting the UID you expect.  Otherwise you could be waiting for a different instance to be deleted.  if UIDs don't match, then you're done


```
pendingFinalization := u.GetDeletionTimestamp() != nil && !u.GetDeletionTimestamp().IsZero()
if pendingFinalization {
  waitingForFinalization := len(resource.UID) > 0
  if waitingForFinalization && resource.UID != u.GetUID{
    // the instance is deleted, so do not add to the resourcesPendingDeletion list
    continue
  }

  // if we aren't waiting for finalization or the UID does match, then we want to ensure we add it to the pending list
  copyResource
  copyResource.UID = u.GetUID
  deletingResources = append(deletingResources, resource)
  continue
}

File: pkg/helper/helpers.go [hunk 28]
```diff
@@ -235,3 +240,91 @@ func UpdateManifestWorkStatus(
...
+
+			// if we aren't waiting for finalization or the UID does match, then we want to ensure we add it to the pending list
+			newResource := copyAppliedResource(resource)
+			newResource.UID = string(u.GetUID())
+			resourcesPendingFinalization = append(resourcesPendingFinalization, newResource)
+			continue
+		}
+
+		// forget this item if the UID does not match the resource UID we previously deleted
+		if len(resource.UID) > 0 && resource.UID != string(u.GetUID()) {
```
deads2k, line 283: nit: this is `waitingForFinalization && resource.UID != string(u.GetUID())`

pkg/spoke/controllers/deletioncontroller/deletion_controller.go
```diff
@@ -96,6 +102,15 @@ func (m *StaleManifestDeletionController) syncManifestWork(ctx context.Context,
...
 	}
 
+	// delete applied resources which are no longer maintained by manifest work
+	noLongerMaintainedResources := findUntrackedResources(manifestWork.Status.AppliedResources, appliedResources)
+	resourcesPendingFinalization, errs := helper.DeleteAppliedResources(noLongerMaintainedResources, m.spokeDynamicClient)
+	if len(errs) != 0 {
+		return utilerrors.NewAggregate(errs)
+	}
+
+	appliedResources = append(appliedResources, resourcesPendingFinalization...)
```
deads2k, line 112: minor: you need to de-dup these lists.

deads2k:
I'd like my two additional comments in a followup please.

/lgtm
/approve

test/integration/work_test.go
```diff
@@ -352,4 +389,88 @@ var _ = ginkgo.Describe("ManifestWork", func() {
 			util.AssertNonexistenceOfResources([]schema.GroupVersionResource{gvrs[3]}, []string{oldServiceAccount.GetNamespace()}, []string{oldServiceAccount.GetName()}, spokeDynamicClient, eventuallyTimeout, eventuallyInterval)
 		})
 	})
+
+	ginkgo.Context("Foreground deletion", func() {
+		var finalizer = "a.b.c/d"
```
pmorie, line 394: One thing here that I will just note since I've seen this become an issue with assumptions like this during testing: any time you find yourself making assumptions like "a.b.c/d" when composing values during tests, ask yourself if there are any field length requirements you may be violating somewhere. I don't think that's happening here. Just a little bit of advice for the future. I have seen assumptions like this during testing hide API errors with field lengths in the past.

pkg/spoke/controllers/deletioncontroller/deletion_controller.go
elgnay: `appliedResources`, the second one on the above line, are built based on ManifestConditions in status. It contains the resources currently maintained by the manifest work. `resourcesPendingFinalization` is a subset of resources which are no longer maintained by manifest work. So I think there is no duplicate item in the concated slice.

PR #27, opened 2020-07-03
Handle unknown kind returned by GuessObjectGroupVersionKind()

elgnay:
To fix https://github.com/open-cluster-management/backlog/issues/3250, 
* Handle unknown kind returned by GuessObjectGroupVersionKind()
* Add apiextensions v1beta1 to scheme by default.

### DIFF
--- a/pkg/helper/helpers.go
+++ b/pkg/helper/helpers.go
@@ -7,14 +7,32 @@ import (
 
 	workv1client "github.com/open-cluster-management/api/client/work/clientset/versioned/typed/work/v1"
 	workapiv1 "github.com/open-cluster-management/api/work/v1"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcehelper"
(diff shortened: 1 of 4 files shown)


pkg/helper/helpers.go
```diff
@@ -328,3 +340,18 @@ func copyAppliedResource(resource workapiv1.AppliedManifestResourceMeta) workapi
 		UID:       resource.UID,
 	}
 }
+
+// GuessObjectGroupVersionKind returns GVK for the passed runtime object.
+func GuessObjectGroupVersionKind(object runtime.Object) (*schema.GroupVersionKind, error) {
+	gvk := resourcehelper.GuessObjectGroupVersionKind(object)
+	if gvk.Kind != "<unknown>" {
```
pmorie: We should leave a comment here explaining this value

pmorie: Should this be a constant?

PR #28, opened 2020-07-06
Implement appliedmanifestwork

qiujian16:


pkg/spoke/controllers/constants.go
```diff
@@ -2,6 +2,10 @@ package controllers
 
 const (
 	// ManifestWorkFinalizer is the name of the finalizer added to manifestworks. It is used to ensure
-	// all maintained resources of a manifestwork are deleted before the manifestwork itself is deleted
+	// related appliedmanifestwork of a manifestwork are deleted before the manifestwork itself is deleted
 	ManifestWorkFinalizer = "cluster.open-cluster-management.io/manifest-work-cleanup"
+	// AppliedManifestWorkFinalizer is the name of the finalizer added to appliedmanifestwork. It is to
+	// ensure all resource relateds to appliedmanifestwork is deleted before appliedmanifestwork itself
+	// is deleted.
+	AppliedManifestWorkFinalizer = "open-cluster-management.io/applied-manifest-work-cleanup"
```
deads2k: scope this to the `work.open-cluster-management.io/applied-manifest-work-cleanup`

pkg/spoke/controllers/finalizercontroller/manifestwork_finalize_controller.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	// If manifestwork is found, check if it is being deleted and has the finalizer.
+	if err == nil {
+		// no work to do until we're deleted
+		if manifestWork.DeletionTimestamp == nil || manifestWork.DeletionTimestamp.IsZero() {
+			return nil
+		}
+
+		// don't do work if the finalizer is not present
```
deads2k: I think that you need to delete the corresponding AppliedManifestWork if the ManifestWork is deleted even if the finalizer is gone.

pkg/spoke/controllers/finalizercontroller/manifestwork_finalize_controller.go
```diff
@@ -0,0 +1,139 @@
...
+	_, err = m.manifestWorkClient.Update(ctx, manifestWork, metav1.UpdateOptions{})
+	if err != nil {
+		return fmt.Errorf("Failed to remove finalizer from ManifestWork %s/%s: %w", manifestWork.Namespace, manifestWork.Name, err)
+	}
+
+	return nil
+}
+
+// CleanAppliedManifestWork deletes appliedmanifestwork and checks if appliedmanifestwork has been cleaned.
+// It returns true only when no related appliedmanifestwork is found, otherwise tries to delete the appliedmanifestwork
```
deads2k: rather than have this method interpret the value, can it return `(appliedManifestWorkPresent bool, err error)` then the higher level can decide how to handle that?

pkg/spoke/controllers/finalizercontroller/manifestwork_finalize_controller.go
```diff
@@ -0,0 +1,139 @@
...
+		WithInformersQueueKeyFunc(helper.AppliedManifestworkQueueKeyFunc(hubHash), appliedManifestWorkInformer.Informer()).
+		WithSync(controller.sync).ToController("ManifestWorkFinalizer", recorder)
+}
+
+func (m *ManifestWorkFinalizeController) sync(ctx context.Context, controllerContext factory.SyncContext) error {
+	manifestWorkName := controllerContext.QueueKey()
+	klog.V(4).Infof("Reconciling ManifestWork %q", manifestWorkName)
+
+	manifestWork, err := m.manifestWorkLister.Get(manifestWorkName)
+	if err != nil && !errors.IsNotFound(err) {
```
deads2k, line 67:
```go
switch{
case errors.IsNotFound(err):
    // fall through to delete because the instance isn't present
case err != nil:
    return err // try again later
case manifestWork.DeletionTimestamp == nil || manifestWork.DeletionTimestamp.IsZero():
    return nil // nothing to do because the instance isn't deleted
default:
    // fall through to delete because the instance is deleted
}
```

pkg/spoke/controllers/finalizercontroller/appliedmanifestwork_finalize_controller.go
```diff
@@ -0,0 +1,135 @@
...
+	if !found {
+		return nil
+	}
+
+	var err error
+
+	// Work is deleting, we remove its related resources on spoke cluster
+	// We still need to run delete for every resource even with ownerref on it, since ownerref does not handle cluster
+	// scoped resource correctly.
+	resourcesPendingFinalization, errs := helper.DeleteAppliedResources(appliedManifestWork.Status.AppliedResources, m.spokeDynamicClient)
```
deads2k, line 99: I'm missing the spot where you're setting appliedmanifestworks.status.appliedresources

deads2k:
I like the idea.  If it is possible, I would split the PR into pieces.

qiujian16:
ok, i will split the pr, that would be easier  to read.

pkg/spoke/controllers/finalizercontroller/appliedmanifestwork_finalize_controller.go
qiujian16: It is in deletioncontroller that handle the status update. Not a good name, i will rename it.

qiujian16:
> ok, i will split the pr, that would be easier to read.

I think it over again, splitting to several pr would make integration test fails..

pkg/spoke/controllers/constants.go
qiujian16: fixed

pkg/spoke/controllers/finalizercontroller/manifestwork_finalize_controller.go
qiujian16: nice!

pkg/spoke/controllers/finalizercontroller/manifestwork_finalize_controller.go
qiujian16: done

pkg/spoke/controllers/finalizercontroller/manifestwork_finalize_controller.go
qiujian16: done

.gitignore
```diff
@@ -4,6 +4,7 @@ work
 *.dll
 *.so
 *.dylib
+!vendor/**
```
deads2k, line 7: just a question, what does this do?

pkg/spoke/controllers/finalizercontroller/manifestwork_finalize_controller.go
```diff
@@ -0,0 +1,143 @@
...
+		WithSync(controller.sync).ToController("ManifestWorkFinalizer", recorder)
+}
+
+func (m *ManifestWorkFinalizeController) sync(ctx context.Context, controllerContext factory.SyncContext) error {
+	manifestWorkName := controllerContext.QueueKey()
+	appliedManifestWorkName := fmt.Sprintf("%s-%s", m.hubHash, manifestWorkName)
+	klog.V(4).Infof("Reconciling ManifestWork %q", manifestWorkName)
+
+	manifestWork, err := m.manifestWorkLister.Get(manifestWorkName)
+	if err != nil && !errors.IsNotFound(err) {
```
deads2k, line 67: minor: this block isn't needed anymore

pkg/spoke/controllers/finalizercontroller/manifestwork_finalize_controller.go
```diff
@@ -0,0 +1,143 @@
...
+
+	return m.appliedManifestWorkClient.Delete(ctx, appliedManifestWorkName, metav1.DeleteOptions{})
+}
+
+// ifAppliedManifestWorkPresent checks appliedmanifestwork is present.
+// It returns false only when appliedmanifestwork is not found, otherwise returns true.
+func (m *ManifestWorkFinalizeController) ifAppliedManifestWorkPresent(ctx context.Context, appliedManifestWorkName string) (bool, error) {
+	_, err := m.appliedManifestWorkLister.Get(appliedManifestWorkName)
+	switch {
+	case errors.IsNotFound(err):
```
deads2k, line 136: minor: we normally do this inline above

pkg/spoke/controllers/finalizercontroller/manifestwork_finalize_controller.go
```diff
@@ -0,0 +1,143 @@
...
+		manifestWorkClient:        manifestWorkClient,
+		manifestWorkLister:        manifestWorkLister,
+		appliedManifestWorkClient: appliedManifestWorkClient,
+		appliedManifestWorkLister: appliedManifestWorkInformer.Lister(),
+		hubHash:                   hubHash,
+		rateLimiter:               workqueue.NewItemExponentialFailureRateLimiter(5*time.Millisecond, 1000*time.Second),
+	}
+
+	return factory.New().
+		WithInformersQueueKeyFunc(func(obj runtime.Object) string {
```
deads2k, line 53: you need to trigger based on appliedmanifestwork changes too. otherwise you can miss a delete of appliedmanifestwork and forget to create a new instance.

deads2k: nm, its the next line down

deads2k:
good code
/lgtm
/approve

/hold

holding in case you have a merge order thing.  Just do the followups separately so this can merge.

PR #35, opened 2020-07-20
Update e2e test cases to work with registration-operator

elgnay:


deploy/spoke/clusterrole.yaml
```diff
@@ -0,0 +1,36 @@
...
+  name: open-cluster-management:work:agent
+rules:
+# Allow agent to get/list/watch/create/delete crds.
+- apiGroups: ["apiextensions.k8s.io"]
+  resources: ["customresourcedefinitions"]
+  verbs: ["get", "list", "watch", "create", "delete", "update"]
+# Allow agent to create/delete namespaces, get/list are contained in admin role already
+- apiGroups: [""]
+  resources: ["namespaces"]
+  verbs: ["create", "delete"]
```
deads2k, line 14: It's unusual to have mutation, but not get privileges.  Intentional?

deploy/spoke/clusterrole.yaml
```diff
@@ -0,0 +1,36 @@
...
+  resources: ["clusterroles", "roles"]
+  verbs: ["get", "list", "watch", "create", "update", "patch", "delete", "escalate", "bind"]
+# Allow agent to create sar
+- apiGroups: ["authorization.k8s.io"]
+  resources: ["subjectaccessreviews"]
+  verbs: ["create"]
+# Allow agent to create events
+- apiGroups: ["", "events.k8s.io"]
+  resources: ["events"]
+  verbs: ["create", "patch", "update"]
```
deads2k: same question about get privileges

deploy/spoke/clusterrole.yaml
```diff
@@ -0,0 +1,36 @@
...
+# Allow agent to managed appliedmanifestworks
+- apiGroups: ["work.open-cluster-management.io"]
+  resources: ["appliedmanifestworks"]
+  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+- apiGroups: ["work.open-cluster-management.io"]
```
deads2k, line 34: see https://github.com/open-cluster-management/registration-operator/pull/49/files#diff-15835a369867a0527a71590c6693ae68R468-R470 . I think you need update on appliedmanifestworks/finalize too.

test/e2e/deployer.go
```diff
@@ -0,0 +1,354 @@
...
+		},
+		Data: data,
+	}
+	_, err = d.spokeKubeClient.CoreV1().Secrets(secret.Namespace).Create(context.TODO(), secret, metav1.CreateOptions{})
+	if err != nil {
+		return err
+	}
+
+	// create cluster role
+	clusterRole, err := assetToUnstructured("deploy/spoke/clusterrole.yaml")
```
deads2k: I would prefer using something something like the `resourceapply.ApplyDirectly` and listing them filenames.  You'd be able to wrap them in the name setter.

deads2k:
/lgtm
/approve
/hold

how tight is your timeframe here.  I've requested some refactoring that I think will help in the end, but I'm willing to take a test that adds protection/reliability now and refactor later if your timeframe is tight.  If that's case, feel free to remove the hold yourself.

deploy/spoke/clusterrole.yaml
elgnay: As the comments above the config says that get/list are contained in admin role already, which will be bound to the same service account.

deploy/spoke/clusterrole.yaml
elgnay: Fixed

deploy/spoke/clusterrole.yaml
elgnay: Fixed

test/e2e/deployer.go
elgnay: Code refactored. After discussion with qiujian16, we decided to use fixed name for resources, like component namespace cluster role and cluster rolebindings, when deploying work agent for e2e testing, because there is no conflicts between two runs of e2e test.

test/e2e/suite_test.go
```diff
@@ -9,20 +10,30 @@ import (
...
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
+const (
+	clusterName = "cluster1"
```
deads2k, line 25: This should be generated in a future PR/.

deads2k:
it's moving forward. I'd like to see the cluster name unique per run in a follow up

/lgtm
/approve

PR #36, opened 2020-07-20
Follow up of pr28

qiujian16:


pkg/spoke/controllers/finalizercontroller/manifestwork_finalize_controller.go
```diff
@@ -86,15 +83,15 @@ func (m *ManifestWorkFinalizeController) sync(ctx context.Context, controllerCon
 		return nil
 	}
 
-	appliedManifestWorkPresent, err := m.ifAppliedManifestWorkPresent(ctx, appliedManifestWorkName)
-	if err != nil {
+	_, err = m.appliedManifestWorkLister.Get(appliedManifestWorkName)
+	switch {
```
deads2k, line 87: optional nit: 

```go
switch{
case errors.IsNotFound(err):
   // if the instance is not found, then we simply continue below this block to remove the finalizer
case err != nil:
  // return all unexpected errors
  return err
default:
  // appliedmanifestwork still exists, requeue the manifestwork to check in the next loop.
}
```

It keeps each condition small and tests every state of err.  It's an optional change.

deads2k:
optional suggestion here: https://github.com/open-cluster-management/work/pull/36/files#r457338622

Up your discretion whether to take it.

/lgtm
/hold

holding in case you want to take the suggestion.

pkg/spoke/controllers/finalizercontroller/manifestwork_finalize_controller.go
qiujian16: yes, it is much clearer. Updated

PR #39, opened 2020-07-31
Add controller to handle Available status

elgnay:


pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+	return nil
+}
+
+func (c *AvailableStatusController) syncManifestWork(ctx context.Context, originalManifestWork *workapiv1.ManifestWork) error {
+	klog.V(4).Infof("Reconciling ManifestWork %q", originalManifestWork.Name)
+	manifestWork := originalManifestWork.DeepCopy()
+
+	// no work if the manifestwork is terminating
+	if !manifestWork.DeletionTimestamp.IsZero() {
```
qiujian16: I think we should still update available status when it is being deleted.

pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
```diff
@@ -0,0 +1,265 @@
...
+	if errors.IsNotFound(err) {
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+func mergeStatusConditions(conditions []workapiv1.StatusCondition, newCondition workapiv1.StatusCondition) []workapiv1.StatusCondition {
```
qiujian16, line 239: what is the difference between this and helpser.MergeStatusConditions?

elgnay: helper.MergeStatusConditions(conditions, newConditions) returns A + B, while:
  - A is the intersection of conditions and newConditions, which is merged
  - B is the difference of newConditions and conditions (newConditions - conditions)

And mergeStatusConditions(conditions, newCondition) returns the union of [newCondition] and conditions, while the intersection of these two slices is merged.

pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
elgnay: Fixed

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -378,7 +378,34 @@ func buildAppliedStatusCondition(err error) workapiv1.StatusCondition {
 	}
 }
 
-func buildManifestResourceMeta(index int, object runtime.Object, restMapper *resource.Mapper) (resourceMeta workapiv1.ManifestResourceMeta, err error) {
+// buildManifestResourceMeta returns resource meta for manifest. It tries to get the resouce
```
pmorie: typo: resource

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -378,7 +378,34 @@ func buildAppliedStatusCondition(err error) workapiv1.StatusCondition {
...
+	errs := []error{}
+
+	resourceMeta, err = buildResourceMeta(index, object, restMapper)
+	if err != nil {
+		errs = append(errs, err)
+	} else if len(resourceMeta.Kind) > 0 && len(resourceMeta.Version) > 0 && len(resourceMeta.Name) > 0 {
+		return resourceMeta, nil
+	}
+
+	// try to get resource meta from menifest if the one got from apply result is incompleted
```
pmorie: typo: manifest

pkg/spoke/controllers/manifestcontroller/manifestwork_controller_test.go
```diff
@@ -562,7 +562,7 @@ func TestAllInCondition(t *testing.T) {
 	}
 }
 
-func TestBuildManifestResourceMeta(t *testing.T) {
+func TestBuildResourceMeta(t *testing.T) {
```
pmorie, line 565: Are you going to add a test case for the new method?

pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
```diff
@@ -0,0 +1,260 @@
...
+	return workapiv1.StatusCondition{
+		Type:    conditionType,
+		Status:  metav1.ConditionFalse,
+		Reason:  "ResourceNotAvailable",
+		Message: "Resource is not available",
+	}
+}
+
+// isResourceAvailable checks if the specific resource is available or not
+func isResourceAvailable(namespace, name string, gvr schema.GroupVersionResource, dynamicClient dynamic.Interface) (bool, error) {
```
pmorie, line 228: If a work has a resource X that has spec A, and then the work is updated to have spec B, is that resource still available?

pmorie: I'm having a hard time understanding how something becomes 'unavailable' until it is wholly removed from the work and the spoke.

pmorie (commented):
Questions about semantics of this condition

pkg/spoke/controllers/manifestcontroller/manifestwork_controller_test.go
elgnay: Yes, I'll add a test case for the new method.

pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
elgnay: > If a work has a resource X that has spec A, and then the work is updated to have spec B, is that resource still available?

If the spec update does not change resource meta (including GVR, namespace and name), the resource will be thought available no matter spec B is applied successfully or not; otherwise it depends on whether spec B is applied successfully.

elgnay: > I'm having a hard time understanding how something becomes 'unavailable' until it is wholly removed from the work and the spoke.

If the controller cannot find the resource with resource meta (including GVR, namespace and name), which is extracted either from apply result of manifest or from manifest itself, the manifest will have an Available condition with status False. It happens when the manifest is failed to apply or the applied resource is deleted by others for some reason.

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
elgnay: Fixed

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
elgnay: Fixed

PR #40, opened 2020-08-17
update websocket and runc lib

skeeey:
see https://github.com/open-cluster-management/backlog/issues/4422

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.13
 require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/go-bindata/go-bindata v3.1.2+incompatible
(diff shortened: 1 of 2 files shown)


pmorie:
Is this still on hold? It seems like it should be fine to me and CI is green.

/approve

skeeey:
@pmorie  we think we don't need to update these libs, these vulnerable libs should be false negative report, see the comments https://github.com/open-cluster-management/backlog/issues/4422#issuecomment-674732404, so after security team confirm this, I will close this pr, :)

pmorie:
@skeeey 

> these vulnerable libs should be false negative report

Ack, looks that way to me. Close when you're ready.

PR #42, opened 2020-08-24
Add appliedmanifestwork webhook

qiujian16:


pkg/webhook/appliedmanifestwork_webhook_test.go
```diff
@@ -0,0 +1,102 @@
...
+				Operation: admissionv1beta1.Create,
+			},
+			hubHash:                 "test",
+			appliedManifestWorkName: "test-work-0",
+			expectedResponse: &admissionv1beta1.AdmissionResponse{
+				Allowed: true,
+			},
+		},
+		{
+			name: "validate creating AppliedManaifestWork with invalid name",
```
pmorie, line 57: typy: Manaif

pkg/webhook/appliedmanifestwork_webhook_test.go
```diff
@@ -0,0 +1,102 @@
...
+				},
+			},
+			hubHash:                 "test",
+			appliedManifestWorkName: "test-work-0",
+			expectedResponse: &admissionv1beta1.AdmissionResponse{
+				Allowed: true,
+			},
+		},
+		{
+			name: "validate creating AppliedManaifestWork",
```
pmorie, line 45: typo

pkg/webhook/appliedmanifestwork_webhook_test.go
```diff
@@ -0,0 +1,102 @@
...
+			hubHash:                 "test",
+			appliedManifestWorkName: "test-work-1",
+			expectedResponse: &admissionv1beta1.AdmissionResponse{
+				Allowed: false,
+				Result: &metav1.Status{
+					Status: metav1.StatusFailure, Code: http.StatusBadRequest, Reason: metav1.StatusReasonBadRequest,
+					Message: "invalid appliedmanifestwork name",
+				},
+			},
+		},
```
pmorie, line 87: would you add a test case for invalid payload?

pmorie (commented):
Looking really good, just a couple notes

deploy/webhook/appliedmanifestwork_webhook.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - operations:
+    - CREATE
+    - UPDATE
+    apiGroups:
+    - work.open-cluster-management.io
+    apiVersions:
+    - "*"
+    resources:
+    - appliedmanifestworks
+  admissionReviewVersions: ["v1beta1"]
```
RainbowMango, line 24: Why not use `v1` for `admissionReviewVersions`?

qiujian16: yes, i think we could use v1 :) will update that

PR #44, opened 2020-08-28
Resolve condition dangling

qiujian16:
1. change the condition merge logic by using existing condition as base and update existing condition using the new condition.
2. update the status update in manifestwork controller and statuscontroller, so we always merge the conditions before each status update.

https://github.com/open-cluster-management/backlog/issues/4819

### DIFF
--- a/pkg/helper/helper_test.go
+++ b/pkg/helper/helper_test.go
@@ -202,10 +202,22 @@ func TestMergeManifestConditions(t *testing.T) {
 				newManifestCondition(0, "resource1", newCondition("one", "True", "my-reason", "my-message", nil)),
 			},
 			newConditions: []workapiv1.ManifestCondition{
-				newManifestCondition(0, "resource1", newCondition("two", "True", "my-reason", "my-message", nil)),
+				newManifestCondition(0, "resource1", newCondition("one", "False", "my-reason", "my-message", nil)),
 			},
 			expectedConditions: []workapiv1.ManifestCondition{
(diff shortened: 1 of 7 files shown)


pmorie:
Nice

pkg/helper/helper_test.go
```diff
@@ -202,10 +202,22 @@ func TestMergeManifestConditions(t *testing.T) {
 				newManifestCondition(0, "resource1", newCondition("one", "True", "my-reason", "my-message", nil)),
 			},
 			newConditions: []workapiv1.ManifestCondition{
-				newManifestCondition(0, "resource1", newCondition("two", "True", "my-reason", "my-message", nil)),
+				newManifestCondition(0, "resource1", newCondition("one", "False", "my-reason", "my-message", nil)),
```
pmorie, line 205: ouch, wish we had caught this before :(

pkg/spoke/controllers/manifestcontroller/manifestwork_controller_test.go
```diff
@@ -348,7 +348,6 @@ func TestFailedToApplyResource(t *testing.T) {
 
 	// Add a reactor on fake client to throw error when creating secret on namespace ns2
 	controller.kubeClient.PrependReactor("create", "secrets", func(action clienttesting.Action) (handled bool, ret runtime.Object, err error) {
-		fmt.Printf("the action get into %v\n", action)
```
pmorie, line 351: nice catch :)

PR #61, opened 2021-02-05
Replace resource mapper with dynamicRESTMapper

elgnay:
The related issue: https://github.com/open-cluster-management/backlog/issues/8944


pkg/spoke/resource/mapper.go
```diff
@@ -35,19 +36,30 @@ func (p *Mapper) Run(stopCh <-chan struct{}) {
 	go wait.Until(func() {
 		p.syncLock.Lock()
 		defer p.syncLock.Unlock()
-		deferredMappd := p.Mapper.(*restmapper.DeferredDiscoveryRESTMapper)
+		deferredMappd := p.StandbyMapper.(*restmapper.DeferredDiscoveryRESTMapper)
 		deferredMappd.Reset()
+		// swap two mapper if the standby mapper is reset and ready to serve
+		if ready(p.StandbyMapper) {
+			p.ServingMapper, p.StandbyMapper = p.StandbyMapper, p.ServingMapper
```
qiujian16: I think we could put lock within this if?

pkg/spoke/resource/mapper.go
```diff
@@ -35,19 +36,30 @@ func (p *Mapper) Run(stopCh <-chan struct{}) {
...
 	if err != nil {
 		return nil, fmt.Errorf("the server doesn't have a resource type %q", gvk.Kind)
 	}
 
 	return mapping, nil
 }
+
+// ready returns true if the given mapper is ready to serve
+func ready(mapper meta.RESTMapper) bool {
+	// if the mapper is able to handle at lease 1 group kind it is ready to serve
```
qiujian16: at least

qiujian16 (commented):
i am also thinking whether NewDynamicRESTMapper is a better way to resolve our problem

pkg/spoke/resource/mapper.go
elgnay: Yes, you are right. We can do it for only `p.ServingMapper` might be accessed in different Goroutines.

pkg/spoke/resource/mapper.go
elgnay: Corrected.

elgnay:
> i am also thinking whether NewDynamicRESTMapper is a better way to resolve our problem

I tried dynamicRESTMapper and it's good in our use case. It caches groupResources and try to refresh when the query does not hit the cache. I've refactored the code to use dynamicRESTMapper.

PR #66, opened 2021-03-16
Ignore version when finding untracked resource

qiujian16:


pkg/spoke/controllers/appliedmanifestcontroller/appliedmanifestwork_controller.go
```diff
@@ -201,19 +201,21 @@ func (m *AppliedManifestWorkController) syncManifestWork(
...
+// API version should be ignored when checking if a resource is no longer tracked by a manifestwork.
+// This is because we treat resources of same GroupResource but different version equivalent.
 func findUntrackedResources(appliedResources, newAppliedResources []workapiv1.AppliedManifestResourceMeta) []workapiv1.AppliedManifestResourceMeta {
 	var untracked []workapiv1.AppliedManifestResourceMeta
 
 	resourceIndex := map[workapiv1.AppliedManifestResourceMeta]struct{}{}
 	for _, resource := range newAppliedResources {
 		key := resource.DeepCopy()
-		key.UID = ""
+		key.Version = ""
```
deads2k: does the `DeleteAppliedResources` still work without version?  It sometimes takes a little fitting to use discovery to find alternative versions.

pkg/spoke/controllers/appliedmanifestcontroller/appliedmanifestwork_controller.go
```diff
@@ -201,19 +201,21 @@ func (m *AppliedManifestWorkController) syncManifestWork(
...
 // findUntrackedResources returns applied resources which are no longer tracked by manifestwork
+// API version should be ignored when checking if a resource is no longer tracked by a manifestwork.
+// This is because we treat resources of same GroupResource but different version equivalent.
 func findUntrackedResources(appliedResources, newAppliedResources []workapiv1.AppliedManifestResourceMeta) []workapiv1.AppliedManifestResourceMeta {
 	var untracked []workapiv1.AppliedManifestResourceMeta
 
 	resourceIndex := map[workapiv1.AppliedManifestResourceMeta]struct{}{}
 	for _, resource := range newAppliedResources {
 		key := resource.DeepCopy()
-		key.UID = ""
```
deads2k, line 210: why was this removed?

pkg/spoke/controllers/appliedmanifestcontroller/appliedmanifestwork_controller.go
qiujian16: we still need version in DeleteAppliedResources. Hence we deepcopy the resource and remove the version only to compare whether the resource should be tracked.

pkg/spoke/controllers/appliedmanifestcontroller/appliedmanifestwork_controller.go
qiujian16: I do not think we need to empty the uid. For the newAppliedResources, we also get its uid by Get before calling findUntrackedResources.  And I think uid will not change when resource version is changed. 

If the uid in newAppliedResource is different than what is recorded in appliemanifestwork, we will regard the recorded one in appliemanifestwork as untracked, and delete the resource with the uid of the recorded one as a precondition https://github.com/open-cluster-management/work/blob/11f530f467683e5b8d6daaff80d32fad97cf54b8/pkg/helper/helpers.go#L211-L218. It should also be fine because we will get an IsNotFound error when delete it.

I will update the doc to explain it.

deads2k: > and delete the resource with the uid of the recorded one as a precondition

does this result in controller fighting during shared resource situations where each one sees a mismatching UID?

pkg/spoke/controllers/appliedmanifestcontroller/appliedmanifestwork_controller.go
deads2k: > we still need version in DeleteAppliedResources. Hence we deepcopy the resource and remove the version only to compare whether the resource should be tracked.

Ok, I can see how that would work.  You could probably improve deletion by having it properly delete from any version, but I can see that as outside the scope of this PR.

pkg/spoke/controllers/appliedmanifestcontroller/appliedmanifestwork_controller.go
```diff
@@ -201,19 +201,26 @@ func (m *AppliedManifestWorkController) syncManifestWork(
 }
 
 // findUntrackedResources returns applied resources which are no longer tracked by manifestwork
+// API version should be ignored when checking if a resource is no longer tracked by a manifestwork.
+// This is because we treat resources of same GroupResource but different version equivalent.
+// It also compares UID of the appliedResources to identify the untracked appliedResources because
+// 1. The UID should keep the same for resources with different versions.
+// 2. The UID in the newAppliedResources is always the latest updated one. The only possibility that UID
+// in appliedResources differs from what in newAppliedResources is that this resource is recreated.
+// Its UID in appliedResources is invalid hence recording it as untracked applied resource and delete it is safe.
```
deads2k, line 210: Doesn't this change the semantics of co-ownership of a resource and resource adoption?  It seems like this would result in controllers that keep trying to delete the shared resource.

pkg/spoke/controllers/appliedmanifestcontroller/appliedmanifestwork_controller.go
qiujian16: If manifestwork/foo and manifestwork/bar have a shared resource,  and we delete manifestwork/foo.
The shared resource will be deleted.  manifestwork/bar will recreate the resource. The current UID of the resource will be mismatched with what is recorded in appliedmanifestwork/bar, so resource with elder uid in appliedmanifestwork/bar is identified as untracked and deleted. The deletion will return a NotFound error and be ignored. Then the current UID of resource will be recorded in appliedmanifestwork/bar. On the other hand, appliedmanifestwork/foo which is in deleting state will know that the newly created resource has a mismatched uid with what is recorded in appliedmanifestwork/foo, and treat the original resource as deleted, so it will not delete the newly created resource.

Mismatch also happens when user creates manifestwork/foo at first, delete it and then creates   manifestwork/bar with a shared resource immediately.  appliedmanifestwork/bar will recreate the resource and appliedmanifestwork/foo will record a mismatched UID. In this case, the resource will be deleted with a mismatched uid, return a NotFound error and the error will be ignored. The newly created resource by manifestwork/bar will not be deleted.

pkg/spoke/controllers/appliedmanifestcontroller/appliedmanifestwork_controller.go
qiujian16: If the resource is shared by manifestwork/foo and manifestwork/bar. Deleting manifestwork/foo will cause resource deleted and recreated once.  It is because we do not update resource uid in deleting appliedmanifestwork. So deleting appliedmanifeswork will not make it to delete the newly created shared resource.

We may consider do not delete the shared resource at all in another enhancement? For instance, a shared resource could have multiple ownerrefs of appliemanifestworks, and if one of the appliedmanifestwork is deleted. We remove the ownerref instead of deleting the resource.

pkg/spoke/controllers/appliedmanifestcontroller/appliedmanifestwork_controller.go
deads2k: > The shared resource will be deleted.

This is only the case until you create the orphaning policy.



> then creates manifestwork/bar with a shared resource immediately. appliedmanifestwork/bar will recreate the resource and appliedmanifestwork/foo will record a mismatched UID. In this case, the resource will be deleted with a mismatched uid, return a NotFound error and the error will be ignored. The newly created resource by manifestwork/bar will not be deleted.

Is adoption still possible?  The case where you create an instance and then later create a matching manifest work.

deads2k:
What if we limit this PR to changing the version matching and add some tests to simulate (or actually)
1. adoption
2. sharing

qiujian16:
@deads2k I've added some tests. Please review again

test/integration/work_test.go
```diff
@@ -554,4 +554,218 @@ var _ = ginkgo.Describe("ManifestWork", func() {
...
+			// Create another manifestworks with one shared resource.
+			// TODO We might not want the sharing in this cases, since the content of a resource in two manifestworks
+			// can be different.
+			anotherWork = util.NewManifestWork(o.SpokeClusterName, "sharing-resource-work", []workapiv1.Manifest{manifests[0]})
+			anotherWork, err = hubWorkClient.WorkV1().ManifestWorks(o.SpokeClusterName).Create(context.Background(), anotherWork, metav1.CreateOptions{})
+			anotherAppliedManifestWorkName = fmt.Sprintf("%s-%s", hubHash, anotherWork.Name)
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		})
+
+		ginkgo.It("shared resource between the manifestwork should be recreated when one manifestwork is deleted", func() {
```
deads2k, line 578: after the orphaning policy is done, this test should be updated to ensure nothing is deleted.  I suspect that's where you'll hit issues

test/integration/work_test.go
```diff
@@ -554,4 +554,218 @@ var _ = ginkgo.Describe("ManifestWork", func() {
 			util.AssertAppliedManifestWorkDeleted(appliedManifestWork.Name, spokeWorkClient, eventuallyTimeout, eventuallyInterval)
 		})
 	})
+
+	ginkgo.Context("Resource sharing and adoption between manifestworks", func() {
```
deads2k, line 558: an adoption workflow requires something  like
1. create configmap
2. create manifestwork with same configmap content
3. ensure the manifestwork does not delete the configmap and recreate it. The existing configmap should be preserved.

deads2k:
The new tests look good.  I think the one I'm worried about about is an adoption workflow requires something like

1. create configmap
2. create manifestwork with same configmap content
3. ensure the manifestwork does not delete the configmap and recreate it. The existing configmap should be preserved.

/lgtm
/approve
/hold

holding for proof of working test.  Once you have proof (maybe in a PR building on this PR?) feel free to drop the hold and merge.

PR #68, opened 2021-04-07
upgrade CRD to v1 and k8s lib to v0.21.0-rc.0

zhiweiyin318:


pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -206,7 +207,7 @@ func (m *ManifestWorkController) applyManifest(manifests []workapiv1.Manifest, r
 	// TODO we should check the certain error.
 	for index, result := range results {
 		// Use dynamic client when scheme cannot decode manifest or typed client cannot handle the object
-		if isDecodeError(result.Error) || isUnhandledError(result.Error) {
+		if isDecodeError(result.Error) || isUnhandledError(result.Error) || isUnsupportedError(result.Error) {
```
qiujian16, line 210: why we handle unsupported?

zhiweiyin318: the latest library-go add Unstructured case into switch to handle `ServiceMonitor` this PR https://github.com/openshift/library-go/pull/1042.

PR #82, opened 2021-06-24
Fix no docker-images load before e2e test

xuezhaojun:
Signed-off-by: xuezhaojun <zxue@redhat.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,7 +61,14 @@ make test-integration
 
 Run e2e testing. During e2e testing, hub and managed cluster share the same Kubernetes cluster.
 ```shell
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -62,6 +62,11 @@ make test-integration
 Run e2e testing. During e2e testing, hub and managed cluster share the same Kubernetes cluster.
 ```shell
 export KUBECONFIG=</path/to/kubeconfig>
+
```
qiujian16, line 67: should we also add go install image-builder here?

xuezhaojun: Of course. 
Added.

Issue #87, opened 2021-07-19
Two version of same resoruce in a manifestwork may result in update conflicts sometime


elgnay:

elgnay:
Closing for it has been fixed.

PR #89, opened 2021-07-20
Resolve apply conflict

qiujian16:


elgnay:
The related issue: https://github.com/open-cluster-management-io/work/issues/87

PR #90, opened 2021-07-30
Support delete option

qiujian16:


elgnay:
This PR is very clear and fine with me.

wonderflow (commented):
How to adopt an orphan resource by creating a new manifest work?

qiujian16:
> How to adopt an orphan resource by creating a new manifest work?

create a manifestwork with manifest pointing to the orphaned resource but no delete option.

wonderflow:
> > How to adopt an orphan resource by creating a new manifest work?
> 
> create a manifestwork with manifest pointing to the orphaned resource but no delete option.

Cool, It would be better if we can have some examples.

qiujian16:
> > > How to adopt an orphan resource by creating a new manifest work?
> > 
> > 
> > create a manifestwork with manifest pointing to the orphaned resource but no delete option.
> 
> Cool, It would be better if we can have some examples.

ok will add an integration test case on this.

pkg/helper/helpers.go
```diff
@@ -198,6 +203,25 @@ func DeleteAppliedResources(resources []workapiv1.AppliedManifestResourceMeta, r
 			continue
 		}
 
+		// If resource is not owned by the appliedManifestWork, ignore it
+		if !IsOwnedBy(owner, u.GetOwnerReferences()) {
+			continue
+		}
+
+		// If there are multiple owners, remove ours and skip
+		remainigOwners := RemoveOwner(owner, u.GetOwnerReferences())
```
deads2k: typo remainingOwners

pkg/helper/helpers.go
```diff
@@ -198,6 +203,25 @@ func DeleteAppliedResources(resources []workapiv1.AppliedManifestResourceMeta, r
...
+		if len(remainigOwners) > 0 {
```
deads2k: nit, factoring like this would have helped me.

```go
if currOwners  := u.GetOwnerReferences(); currOwners > 1{
  remainingOwners := RemoveOwner(owner, u.GetOwnerReferences())

  // do stuff
  continue
}
```

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -194,71 +201,87 @@ func (m *ManifestWorkController) sync(ctx context.Context, controllerContext fac
...
+		return result
+	}
+
 	results := resourceapply.ApplyDirectly(clientHolder, recorder, func(name string) ([]byte, error) {
-		unstructuredObj := &unstructured.Unstructured{}
-		err := unstructuredObj.UnmarshalJSON(manifest.Raw)
-		if err != nil {
-			return nil, err
-		}
-		unstructuredObj.SetOwnerReferences([]metav1.OwnerReference{owner})
```
deads2k, line 227: why are we no longer setting an ownerref during the first apply?

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -194,71 +201,87 @@ func (m *ManifestWorkController) sync(ctx context.Context, controllerContext fac
...
-		result.Result, result.Changed, result.Error = m.applyUnstructrued(manifest.Raw, owner, recorder)
+		result.Result, result.Changed, result.Error = m.applyUnstructrued(manifest.Raw, nil, gvr, recorder)
 	}
 
+	if result.Error != nil {
+		return result
+	}
+
+	//  Apply with unstructure again to update owners
+	result.Error = m.applyOwner(gvr, resMeta.Namespace, resMeta.Name, deleteOption, owner, recorder)
```
deads2k: why are we doing this separate from the "normal" apply?

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -286,6 +309,90 @@ func (m *ManifestWorkController) applyUnstructrued(data []byte, owner metav1.Own
...
+		return err
+	}
+
+	recorder.Eventf(fmt.Sprintf(
+		"Owner of %s Updated", existing.GetKind()), "Updated %s/%s", existing.GetNamespace(), existing.GetName())
+
+	return nil
+}
+
+func manageOwnerRef(
```
deads2k, line 341: It seems like this logic could be called during our normal create flow instead of tripling our number of calls.

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -194,71 +201,87 @@ func (m *ManifestWorkController) sync(ctx context.Context, controllerContext fac
...
+	result := applyResult{}
+
+	resMeta, gvr, err := buildManifestResourceMeta(index, manifest, m.restMapper)
+	result.resourceMeta = resMeta
+	if err != nil {
+		result.Error = err
+		return result
+	}
+
 	results := resourceapply.ApplyDirectly(clientHolder, recorder, func(name string) ([]byte, error) {
```
deads2k: the ApplyDirectly doesn't directly manage ownerReferences?  Should it merge by default, similar to the handling for labels and annotations?

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -286,6 +309,90 @@ func (m *ManifestWorkController) applyUnstructrued(data []byte, owner metav1.Own
...
+	}
+
+	owners, _ := manageOwnerRef(gvr, namespace, name, deleteOption, existing.GetOwnerReferences(), owner)
+
+	if equality.Semantic.DeepEqual(owners, existing.GetOwnerReferences()) {
+		return nil
+	}
+
+	existing.SetOwnerReferences(owners)
+	_, err = m.spokeDynamicClient.
```
deads2k: I would only expect the secondary update for a removal flow if the returned value for the apply indicates a spurious ownerref.

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -429,31 +537,34 @@ func buildManifestResourceMeta(index int, object runtime.Object, manifest workap
...
 
-	return resourceMeta, utilerrors.NewAggregate(errs)
+	return resourceMeta, gvr, utilerrors.NewAggregate(errs)
 }
 
-func buildResourceMeta(index int, object runtime.Object, restMapper meta.RESTMapper) (resourceMeta workapiv1.ManifestResourceMeta, err error) {
+func buildResourceMeta(
+	index int,
+	object runtime.Object,
+	restMapper meta.RESTMapper) (resourceMeta workapiv1.ManifestResourceMeta, gvr schema.GroupVersionResource, err error) {
```
deads2k: past me should have caught this.  Can we eliminate the named return values for this method?

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
qiujian16: ApplyDirectly does not merge ownerReferences today, https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourcemerge/object_merger.go#L12.

We discussed this before that we might add a customized hooks for all apply func, but I haven't done that yet.  I will create a PR in library-go to add that...

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
qiujian16: We can do that with create, but not  update. Since applyDirectly does not merge OwnerRefs today. In that way the flow will be like:

1. Get resources at first to get current owners.
2. MergeOwners in normal apply
3. Get resources again to check if owners are correct.
4. Update owners if it is not correct.

pkg/helper/helpers.go
```diff
@@ -198,6 +208,34 @@ func DeleteAppliedResources(resources []workapiv1.AppliedManifestResourceMeta, r
...
+		if !IsOwnedBy(owner, existingOwner) {
+			continue
+		}
+
+		// Merge with the existing owners to move the owner.
+		modified := resourcemerge.BoolPtr(false)
+		resourcemerge.MergeOwnerRefs(modified, &existingOwner, []metav1.OwnerReference{*ownerCopy})
+
+		// If there are still existing owners, update ownerrefs only.
+		if len(existingOwner) > 0 {
```
deads2k, line 223: Do you want the behavior to be "any other owner" or to be "any other manifestWork  owner"?

If you could update the comment to indicate an explicit choice.

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -196,98 +200,196 @@ func (m *ManifestWorkController) sync(ctx context.Context, controllerContext fac
...
+		return nil, fmt.Errorf("Failed to find gvr from restmapping: %w", err)
 	}
 
-	return mapping.Resource, unstructuredObj, nil
+	return unstructuredObj, nil
 }
 
-func (m *ManifestWorkController) applyUnstructrued(ctx context.Context, data []byte, owner metav1.OwnerReference, recorder events.Recorder) (*unstructured.Unstructured, bool, error) {
-	gvr, required, err := m.decodeUnstructured(data)
+func (m *ManifestWorkController) applyUnstructrued(
```
deads2k: typo unstructured

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -196,98 +200,196 @@ func (m *ManifestWorkController) sync(ctx context.Context, controllerContext fac
...
 	actual, err := m.spokeDynamicClient.Resource(gvr).Namespace(required.GetNamespace()).Update(
-		context.TODO(), required, metav1.UpdateOptions{})
+		ctx, required, metav1.UpdateOptions{})
 	recorder.Eventf(fmt.Sprintf(
 		"%s Updated", required.GetKind()), "Updated %s/%s", required.GetNamespace(), required.GetName())
 	return actual, true, err
 }
 
+// manageOwnerRef update ownerref based on the resource and the deleteOption. If the resource is orphaned.
+// The owner's UID is updated to indicate that the owner should be removed from the existing ownerrefs in the resource.
```
deads2k: it actually returns a reference indicating that the owner should be removed, right?

deads2k:
minor comments.

/lgtm
/hold

holding for minor comments and/or other reviewers.

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
qiujian16: yes. I will update the comment.

pkg/helper/helpers.go
qiujian16: I think it is any other owner, not only manifestWork. I will update the comment.

PR #92, opened 2021-08-09
Install deps for image builder

qiujian16:


zhujian7:
I think we do not need this, the upstream https://github.com/openshift/imagebuilder/issues/199#issuecomment-895571356 fixed it.

Issue #83, opened 2021-06-30
Request that manifestWork conditions carry the `ObservedGeneration` of the ManifestWork


ShyamsundarR:
At the hub, we create and later update certain ManifestWork CRs. Our reconciler hence needs to check if the latest generation of the ManifestWork is applied, before progressing on its tasks.

Without the `ObserveredGeneration` it is not possible to parse if the conditions are stale and hence for the older generation of the work CR or is for the most recent generation.

As the current implementation of ManifestWork does not report `ObserverGeneration` in its conditions, the request is to add the same.

qiujian16:
good catch! we will queue this as an enhancement.

/assign @qiujian16

qiujian16:
it is done

PR #95, opened 2021-08-30
Add generation in status

qiujian16:
https://github.com/open-cluster-management-io/work/issues/83


elgnay:
The change looks good to me.

PR #98, opened 2021-09-08
reduce the numbe of GET request of manifestwork

elgnay:
The related issue: https://github.com/open-cluster-management/backlog/issues/14616


qiujian16:
/approve
/lgtm

should we do the same thing for registration?

PR #101, opened 2021-10-14
Set default propagation policy to background

qiujian16:


qiujian16:
relates to https://github.com/open-cluster-management-io/community/issues/95

PR #104, opened 2021-11-04
support run work agent outside of managed cluster

zhujian7:

related issue: https://github.com/open-cluster-management-io/registration-operator/issues/158

---
todo:
- [ ] add description and script to deploy work agent in Hosted mode. (not contained in this PR)

### DIFF
--- a/pkg/spoke/spokeagent.go
+++ b/pkg/spoke/spokeagent.go
@@ -2,6 +2,7 @@ package spoke
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"open-cluster-management.io/work/pkg/helper"
@@ -16,6 +17,7 @@ import (
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 
 	workclientset "open-cluster-management.io/api/client/work/clientset/versioned"
@@ -24,10 +26,12 @@ import (
 
 // WorkloadAgentOptions defines the flags for workload agent
 type WorkloadAgentOptions struct {
-	HubKubeconfigFile string
-	SpokeClusterName  string
-	QPS               float32
-	Burst             int
+	HubKubeconfigFile   string
+	SpokeKubeconfigFile string
+	SpokeKubeconfig     *rest.Config
+	SpokeClusterName    string
+	QPS                 float32
+	Burst               int
 }
 
 // NewWorkloadAgentOptions returns the flags with default value set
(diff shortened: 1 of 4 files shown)


pkg/spoke/spokeagent.go
```diff
@@ -43,6 +46,7 @@ func (o *WorkloadAgentOptions) AddFlags(cmd *cobra.Command) {
 	flags := cmd.Flags()
 	// This command only supports reading from config
 	flags.StringVar(&o.HubKubeconfigFile, "hub-kubeconfig", o.HubKubeconfigFile, "Location of kubeconfig file to connect to hub cluster.")
+	flags.StringVar(&o.SpokeKubeconfigFile, "spoke-kubeconfig", o.SpokeKubeconfigFile, "Location of kubeconfig file to connect to spoke cluster.")
```
qiujian16: can't we use the built-in --kubeconfig flag here?

zhujian7: I  think it is better not, the built-in `--kubeconfig` in our case represents the cluster where the work-agent pod run.

pkg/spoke/spokeagent.go
```diff
@@ -24,10 +26,12 @@ import (
 
 // WorkloadAgentOptions defines the flags for workload agent
 type WorkloadAgentOptions struct {
-	HubKubeconfigFile string
-	SpokeClusterName  string
-	QPS               float32
-	Burst             int
+	HubKubeconfigFile   string
+	SpokeKubeconfigFile string
+	SpokeKubeconfig     *rest.Config
```
qiujian16: why do we need this?

pkg/spoke/spokeagent.go
```diff
@@ -64,28 +69,39 @@ func (o *WorkloadAgentOptions) RunWorkloadAgent(ctx context.Context, controllerC
...
 	workInformerFactory := workinformers.NewSharedInformerFactoryWithOptions(hubWorkClient, 5*time.Minute, workinformers.WithNamespace(o.SpokeClusterName))
 
-	// Build dynamic client and informer for spoke cluster
-	spokeRestConfig := controllerContext.KubeConfig
-	spokeRestConfig.QPS = o.QPS
-	spokeRestConfig.Burst = o.Burst
-	spokeDynamicClient, err := dynamic.NewForConfig(spokeRestConfig)
+	// load spoke client config and create spoke clients,
+	// the work agent may running not in the spoke/managed cluster.
+	if o.SpokeKubeconfig == nil {
```
qiujian16: I do not see how it can be not nil

pkg/spoke/spokeagent.go
qiujian16: this should be explicit callout that if this is not set, but --kubeconfig is set. kubeconfig will be used to connect to spoke.

pkg/spoke/spokeagent.go
```diff
@@ -64,28 +69,39 @@ func (o *WorkloadAgentOptions) RunWorkloadAgent(ctx context.Context, controllerC
...
+		if o.SpokeKubeconfigFile != "" {
+			o.SpokeKubeconfig, err = clientcmd.BuildConfigFromFlags("" /* leave masterurl as empty */, o.SpokeKubeconfigFile)
+			if err != nil {
+				return fmt.Errorf("unable to load spoke kubeconfig from file %q: %w", o.SpokeKubeconfigFile, err)
```
qiujian16: I would make it a func

```go
 func (o *WorkloadAgentOptions) spokeKubeConfig(controllerContext) *rest.Config, error {
    if o.SpokeKubeconfigFile != "" {
        return clientcmd.BuildConfigFromFlags("" /* leave masterurl as empty */, o.SpokeKubeconfigFile)
    }

   return controllerContext.KubeConfig
}
```

pkg/spoke/spokeagent.go
zhujian7: This was originally used for integration test, but now I think it can be deleted

qiujian16 (commented):
/approve
/lgtm

PR #106, opened 2021-12-02
fix flaky test

zhujian7:

### DIFF
--- a/test/integration/work_test.go
+++ b/test/integration/work_test.go
@@ -801,7 +801,6 @@ var _ = ginkgo.Describe("ManifestWork", func() {
 			// Ensure ownership of configmap is updated
 			gomega.Eventually(func() error {
 				cm, err := spokeKubeClient.CoreV1().ConfigMaps(o.SpokeClusterName).Get(context.Background(), "cm1", metav1.GetOptions{})
-
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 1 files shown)


test/integration/work_test.go
```diff
@@ -949,21 +937,18 @@ var _ = ginkgo.Describe("ManifestWork", func() {
...
 				work, err = hubWorkClient.WorkV1().ManifestWorks(o.SpokeClusterName).Get(context.Background(), work.Name, metav1.GetOptions{})
-
 				if err != nil {
 					return err
 				}
 
 				work.Spec.DeleteOption = nil
-				work, err = hubWorkClient.WorkV1().ManifestWorks(o.SpokeClusterName).Update(context.Background(), work, metav1.UpdateOptions{})
-
+				_, err = hubWorkClient.WorkV1().ManifestWorks(o.SpokeClusterName).Update(context.Background(), work, metav1.UpdateOptions{})
```
zhujian7, line 945: if the `update` failed, we should not assign the value of `work`, otherwise, the work will be `nil`, and the `get`([line 939](https://github.com/open-cluster-management-io/work/pull/106/files#diff-22c00ae1580f36ac14f08b21b95d796167289621703bd0c9a4bd222d8129770bR939)) will be failed in the next **retry** since the `work.Name` is empty.

qiujian16 (commented):
/approve
/lgtm

PR #107, opened 2021-12-20
Add interested field

qiujian16:


pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
```diff
@@ -103,12 +105,23 @@ func (c *AvailableStatusController) syncManifestWork(ctx context.Context, origin
...
 		newConditions := helper.MergeStatusConditions(manifest.Conditions, []metav1.Condition{availableStatusCondition})
-		if !reflect.DeepEqual(manifestWork.Status.ResourceStatus.Manifests[index].Conditions, newConditions) {
+		if !equality.Semantic.DeepEqual(manifestWork.Status.ResourceStatus.Manifests[index].Conditions, newConditions) {
 			manifestWork.Status.ResourceStatus.Manifests[index].Conditions = newConditions
 			needStatusUpdate = true
 		}
+
+		values, err := c.getFeedbackValues(manifest.ResourceMeta, obj, manifestWork.Spec.ManifestConfigs)
+		if err != nil {
+			//TODO Update condition heres
```
yue9944882: bypassing the error will lead to unexpected value update. in the first stage, we should at least return the error and additionally either reflect the error in the conditions or send an event

pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
```diff
@@ -103,12 +105,23 @@ func (c *AvailableStatusController) syncManifestWork(ctx context.Context, origin
 	needStatusUpdate := false
 	// handle status condition of manifests
 	for index, manifest := range manifestWork.Status.ResourceStatus.Manifests {
```
yue9944882, line 111: nit: `manifestWork.Status.ResourceStatus.Manifests` is an array of value (not references), consider avoid direct usage of `manifest` in case of racy condition

pkg/spoke/statusfeedback/reader.go
```diff
@@ -0,0 +1,121 @@
...
+	if len(results) == 0 || len(results[0]) == 0 {
+		// no results are found here.
+		return nil, nil
+	}
+
+	// as we only support simple JSON path, we can assume to have only one result (or none, filtered out above)
+	value := results[0][0].Interface()
+	var fieldValue workapiv1.FieldValue
+	switch t := value.(type) {
+	case int64:
```
yue9944882, line 97: the int64 is converted to 32 implicitly here. i remember it's due to the limitation of unstructured json codec, it's taking all the numeric values to int64, is it? fwiw i think we should assert the overflowing here defensively here such as `int64(int32(t)) == int64(t)`?

pkg/spoke/statusfeedback/reader.go
```diff
@@ -0,0 +1,121 @@
...
+			values = append(values, *value)
+		}
+	}
+
+	return values, utilerrors.NewAggregate(errs)
+}
+
+func getValueByJsonPath(name, path string, obj *unstructured.Unstructured) (*workapiv1.FeedbackValue, error) {
+	j := jsonpath.New(name).AllowMissingKeys(true)
+	err := j.Parse(fmt.Sprintf("{.status%s}", path))
```
yue9944882: implicitly adding `.status` as the prefix of the json path is suspicious. even tho it's always recommended to extend the status information under the `.status` struct but there can be other forms of saving status in the object such as annotations or even the spec.

pkg/spoke/statusfeedback/reader.go
```diff
@@ -0,0 +1,121 @@
...
+		return nil, err
+	}
+
+	results, err := j.FindResults(obj.UnstructuredContent())
+
+	if err != nil {
+		return nil, err
+	}
+
+	if len(results) == 0 || len(results[0]) == 0 {
```
yue9944882, line 82: what if a `[][]interface{} = [[null]]` result? (iiuc the current implementation will make the controller constantly retrying in failure). i think it should be identical to a `[[]]` b/c mostly the only difference between them is whether the go struct has an `omitempty` in the definition. 

and i can think we should add more tests to pin the behavior to enumerate how different feedback data types will be dealt with, including the failure path. ps we can do the testing in the follow-ups.

yue9944882 (commented):
thanks! overall i think we should cover this awesome feature w more unit-cases and e2e-tests.

yue9944882 (commented):
another thing i'm worried about is the bursty client traffic from the dynamic client, especially when the controller is bootstrapping the client will be basically iterating over all the resources managed by the work. i think we should apply a dedicated client-side muzzler for the dynamic client. e.g. the muzzler can limit the maximum request concurrency to a reasonable number such as 5?

pkg/spoke/statusfeedback/reader.go
qiujian16: yes, that is the limitation on json codec. If we assert overflowing here, it will cause that an int64 value cannot be returned, like observedGeneration. I am thinking we should define this as int64 in API. WDYT?

pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
qiujian16: good point. will add a todo here. I think it is the issue of this whole controller on potential race.

pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
qiujian16: return error will fallback and requeue. However, since it is most likely a user setting error, retry will still return error..So I think adding a condition only makes more sense.

qiujian16:
> another thing i'm worried about is the bursty client traffic from the dynamic client, especially when the controller is bootstrapping the client will be basically iterating over all the resources managed by the work. i think we should apply a dedicated client-side muzzler for the dynamic client. e.g. the muzzler can limit the maximum request concurrency to a reasonable number such as 5?

this might have some tradeoffs, since this client is also used to apply resources. A small QPS may cause long time for the resource to be applied. I think we currently set the qps to 100.

pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
```diff
@@ -102,13 +104,25 @@ func (c *AvailableStatusController) syncManifestWork(ctx context.Context, origin
...
-		availableStatusCondition := buildAvailableStatusCondition(manifest.ResourceMeta, c.spokeDynamicClient)
+		obj, availableStatusCondition := buildAvailableStatusCondition(manifest.ResourceMeta, c.spokeDynamicClient)
 		newConditions := helper.MergeStatusConditions(manifest.Conditions, []metav1.Condition{availableStatusCondition})
-		if !reflect.DeepEqual(manifestWork.Status.ResourceStatus.Manifests[index].Conditions, newConditions) {
+		if !equality.Semantic.DeepEqual(manifestWork.Status.ResourceStatus.Manifests[index].Conditions, newConditions) {
 			manifestWork.Status.ResourceStatus.Manifests[index].Conditions = newConditions
 			needStatusUpdate = true
 		}
+
+		values, err := c.getFeedbackValues(manifest.ResourceMeta, obj, manifestWork.Spec.ManifestConfigs)
```
deads2k: I gather that these logically distinct controllers are combined because of the cost of the live-get?  If this is the case, the reasoning should be commented for future readers.

pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
deads2k: `utilruntime.HandleError` is more appropriate for errors that are skipped.

EDIT:
having now read down, I agree with 
> return error will fallback and requeue. However, since it is most likely a user setting error, retry will still return error..So I think adding a condition only makes more sense.

And the only way that a client can know this is to set a condition.

pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
```diff
@@ -129,7 +143,7 @@ func (c *AvailableStatusController) syncManifestWork(ctx context.Context, origin
 	manifestWork.Status.Conditions = workStatusConditions
 
 	// no work if the status of manifestwork does not change
-	if !needStatusUpdate && reflect.DeepEqual(originalManifestWork.Status.Conditions, manifestWork.Status.Conditions) {
+	if !needStatusUpdate && equality.Semantic.DeepEqual(originalManifestWork.Status.Conditions, manifestWork.Status.Conditions) {
```
deads2k: why not do a deep equal on both parts of the status you're updating here instead of computing it above in `needStatusUpdate`?

pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
```diff
@@ -102,13 +104,25 @@ func (c *AvailableStatusController) syncManifestWork(ctx context.Context, origin
...
+
+		values, err := c.getFeedbackValues(manifest.ResourceMeta, obj, manifestWork.Spec.ManifestConfigs)
+		if err != nil {
+			//TODO Update condition heres
+			klog.Errorf("Some values cannot be obtained %v", err)
+		}
+
+		if !equality.Semantic.DeepEqual(manifestWork.Status.ResourceStatus.Manifests[index].StatusFeedbacks.Values, values) {
+			manifestWork.Status.ResourceStatus.Manifests[index].StatusFeedbacks.Values = values
+			needStatusUpdate = true
```
deads2k: minor: rename to `feedbackValuesChanged`?

pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
```diff
@@ -187,58 +201,88 @@ func aggregateManifestConditions(generation int64, manifests []workapiv1.Manifes
 	}
 }
 
+func (c *AvailableStatusController) getFeedbackValues(
+	resourceMeta workapiv1.ManifestResourceMeta, obj *unstructured.Unstructured, manifestOptions []workapiv1.ManifestConfigOption) ([]workapiv1.FeedbackValue, error) {
```
deads2k: obj can be nil when it is not found.   should this method shortcircuit on that?

pkg/spoke/statusfeedback/rules/rule.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+type DefaultWellKnownStatusResolver struct {
+	rules map[schema.GroupKind][]workapiv1.JsonPath
+}
+
+var deploymentRule = []workapiv1.JsonPath{
+	{
+		Name: "ReadyReplicas",
+		Path: ".status.readyReplicas",
```
deads2k, line 19: this path is logically dependent on the version.  The request to get status from an object in the API is logically *not* dependent on version (groupKind), but by the time you get the point of selecting what jsonpath to use, you have the group*version*kind and need to use that.

pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
qiujian16: a comment on the reason is added at the controller definition.

pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
qiujian16: good point, added a check before calling this function

pkg/spoke/statusfeedback/rules/rule.go
qiujian16: right, changed to use GVK as the key here

pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
qiujian16: I update the logic here so needStatusUpdate is not necessary

pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
yue9944882: btw in the legacy kubefed we had a multi-cluster informer that dynamically watches resources by api discovery. if we can replace the live-read w/ informer-read, the cost will be significantly reduced.

yue9944882:
> this might have some tradeoffs, since this client is also used to apply resources. A small QPS may cause long time for the resource to be applied. I think we currently set the qps to 100.

for recent k8s releases, we can also police the api traffic from work agent by setting proper APF rules at the server side. but i think it's reasonable to apply an upper-bound for the emitting requests in case of extreme buggy behaviors defensively, a loose rate-limiter like 100 qps makes sense to me.

pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
qiujian16: set a condition to reflect this

pkg/spoke/statusfeedback/reader.go
```diff
@@ -0,0 +1,127 @@
...
+	case workapiv1.WellKnownStatusType:
+		paths := s.wellKnownStatus.GetPathsByKind(obj.GroupVersionKind())
+		if len(paths) == 0 {
+			return values, fmt.Errorf("cannot find the wellknown statuses for resrouce with gvk %s", obj.GroupVersionKind().String())
+		}
+
+		for _, path := range paths {
+			value, err := getValueByJsonPath(path.Name, path.Path, obj)
+			if err != nil {
+				errs = append(errs, err)
```
deads2k, line 38: you need to be sure that these errors include the `path.Name`, so you'll be able to debug.

deads2k:
one minor comment, lgtm otherwise.

pkg/spoke/controllers/statuscontroller/availablestatus_controller.go
qiujian16: yeah, there is good/bad to have such informers. given the current informer filtering mechanism. We might monitor a lot of unrelated resource on the spoke. It would be quite costly on a large spoke I think.

yue9944882 (commented):
LGTM

PR #117, opened 2022-02-22
Add an integration test to remove a resource with orphanDelete

qiujian16:


qiujian16:
@zhujian7 it seems it works...

PR #119, opened 2022-02-28
[WIP] Add cache for work apply

qiujian16:
1. cache based on manifest hash/resource version/generation
2. reduce the update call

### DIFF
--- a/pkg/helper/cache.go
+++ b/pkg/helper/cache.go
@@ -1,0 +1,226 @@
+package helper
+
+import (
+	"crypto/md5"
+	"fmt"
+	"io"
+	"reflect"
+	"sync"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/klog/v2"
+)
+
+// Learned from library-go but add generation
+
+type cachedVersionKey struct {
+	name      string
+	namespace string
+	kind      schema.GroupKind
+}
+
+// record of resource metadata used to determine if its safe to return early from an ApplyFoo
+// resourceHash is an ms5 hash of the required in an ApplyFoo that is computed in case the input changes
+// resourceVersion is the received resourceVersion from the apiserver in response to an update that is comparable to the GET
+type cachedResource struct {
+	resourceHash, resourceVersion string
+	generation                    int64
+}
+
+type WorkResourceCache struct {
+	cache  map[cachedVersionKey]cachedResource
+	rwLock sync.RWMutex
+}
+
+func NewWorkResourceCache() *WorkResourceCache {
+	return &WorkResourceCache{
+		cache: map[cachedVersionKey]cachedResource{},
+	}
+}
+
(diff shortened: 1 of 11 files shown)


pkg/helper/cache.go
```diff
@@ -0,0 +1,226 @@
...
+
+type cachedVersionKey struct {
+	name      string
+	namespace string
+	kind      schema.GroupKind
+}
+
+// record of resource metadata used to determine if its safe to return early from an ApplyFoo
+// resourceHash is an ms5 hash of the required in an ApplyFoo that is computed in case the input changes
+// resourceVersion is the received resourceVersion from the apiserver in response to an update that is comparable to the GET
```
deads2k, line 27: explain the reason for the generation please.

pkg/helper/cache.go
```diff
@@ -0,0 +1,226 @@
...
+type cachedResource struct {
+	resourceHash, resourceVersion string
+	generation                    int64
```
deads2k, line 32: I think I get it.  If the generation  hasn't changed, then the intent of the object has not changed and therefore you don't need to update it even if the resource version changes.

pkg/helper/cache.go
```diff
@@ -0,0 +1,226 @@
...
+		if versionMatch && hashMatch {
+			klog.V(4).Infof("found matching resourceVersion & manifest hash")
+			return true
+		}
+	}
+
+	return false
+}
+
+func (c *WorkResourceCache) SafeToSkipApplyWithGeneration(required runtime.Object, existing runtime.Object) bool {
```
deads2k, line 187: This needs godoc explaining that it is only safe for cases where
1. the generation is used
2. the generation is honored for the fields you're concerned about
3. how to figure out if generation does what you want

pkg/helper/cache.go
```diff
@@ -0,0 +1,226 @@
...
+		if versionMatch && hashMatch {
+			klog.V(4).Infof("found matching generation & manifest hash")
+			return true
+		}
+	}
+
+	return false
+}
+
+func (c *WorkResourceCache) RemoveCache(required runtime.Object) {
```
deads2k, line 227: this fix seems worthwhile for library-go as well.

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -327,13 +306,20 @@ func (m *ManifestWorkController) applyUnstructured(
 	// Always overwrite required ownerrefs from existing, since ownerrefs of required has been merged to existing
 	required.SetOwnerReferences(existingOwners)
 
+	// Always use generation to check skip for unstructured. It should be fine since resources which do not increment generation
```
deads2k: I'm not clear on why this is true.  I don't know the CR generation incrementing logic offhand.

qiujian16: I think generation of CR always get incremented? https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/strategy.go#L132

PR #128, opened 2022-04-01
Keep the finalizers unchanged for unstructured

zhujian7:


zhujian7:
cc @qiujian16

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -345,6 +347,43 @@ func (m *ManifestWorkController) applyUnstructured(
 	return actual, true, err
 }
 
+// mergeFinalizers merge finalizer with different elements. the caller needs to receive the return value.
+func mergeFinalizers(modified *bool, existing []string, required []string) []string {
```
qiujian16: To be consistent with other func, I think it could be 
```
func mergeStringSlice(modified *bool, existing *[]string, required []string)
```

UPDATE: to be consistent with applyDirectly, I think we should set the existing instead of merge.

qiujian16:
please also add an integration test.

qiujian16 (commented):
/lgtm

PR #139, opened 2022-06-02
support update strategy in manifestwork

qiujian16:


pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -271,12 +293,57 @@ func (m *ManifestWorkController) applyOneManifest(
...
+	}
+
+	existing, err := m.spokeDynamicClient.
+		Resource(gvr).
+		Namespace(required.GetNamespace()).
+		Get(ctx, required.GetName(), metav1.GetOptions{})
+
+	switch {
+	case err == nil:
+		// merge ownerref by existing
```
deads2k: I didn't expect server-side apply to need this.

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -271,12 +293,57 @@ func (m *ManifestWorkController) applyOneManifest(
...
+	gvr schema.GroupVersionResource,
+	required *unstructured.Unstructured,
+	config *workapiv1.ServerSideApplyConfig,
+	recorder events.Recorder) (*unstructured.Unstructured, bool, error) {
+	force := false
+	fieldManager := "work-agent"
+
+	if config != nil {
+		force = config.Force
+		fieldManager = config.FieldManager
```
deads2k: length check?

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -271,12 +293,57 @@ func (m *ManifestWorkController) applyOneManifest(
...
+
+	patch, err := json.Marshal(resourcemerge.WithCleanLabelsAndAnnotations(required))
+	if err != nil {
+		return nil, false, err
+	}
+
+	actual, err := m.spokeDynamicClient.
+		Resource(gvr).
+		Namespace(required.GetNamespace()).
+		Patch(ctx, required.GetName(), types.ApplyPatchType, patch, metav1.PatchOptions{FieldManager: fieldManager, Force: pointer.Bool(force)})
```
deads2k, line 354: Does this work?  I'd be very happy to see a k/k patch to add Apply to the dynamic client.

deads2k:
Minor comments.  The change looks good.

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
qiujian16: Apply is in the client-go's latest code, but not in 2.4. Will add a TODO to change this to Apply

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
qiujian16: I think this is needed when user changes the deleteOption to Orphan. We should remove the ownerref of ApplieManifestWork on this resource. Added a comment to explain this.

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -271,12 +293,61 @@ func (m *ManifestWorkController) applyOneManifest(
...
+		return nil, false, err
+	}
+
+	// TODO use Apply method instead when upgrading the client-go to 0.25.x
+	actual, err := m.spokeDynamicClient.
+		Resource(gvr).
+		Namespace(required.GetNamespace()).
+		Patch(ctx, required.GetName(), types.ApplyPatchType, patch, metav1.PatchOptions{FieldManager: fieldManager, Force: pointer.Bool(force)})
+	recorder.Eventf(fmt.Sprintf(
+		"%s Server Side Applied", required.GetKind()), "Patched %s/%s with field manager %q", required.GetNamespace(), required.GetName(), fieldManager)
```
yue9944882: nit: extra slash when applying non-namespaced resoruces

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -271,12 +293,61 @@ func (m *ManifestWorkController) applyOneManifest(
...
+	fieldManager := "work-agent"
+
+	if config != nil {
+		force = config.Force
+		if len(config.FieldManager) > 0 {
+			fieldManager = config.FieldManager
+		}
+	}
+
+	existing, err := m.spokeDynamicClient.
```
yue9944882: under SSA mode, we are not supposed to do any client-side get any more. all the owner-ref and label/annotation merging should be done at the server-side. can you elaborate the purpose of this?

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -271,12 +293,61 @@ func (m *ManifestWorkController) applyOneManifest(
...
+		Namespace(required.GetNamespace()).
+		Get(ctx, required.GetName(), metav1.GetOptions{})
+
+	switch {
+	case err == nil:
+		// merge ownerref by existing. We need this to remove the ownerref of appliedManifestWork when user
+		// changes the deleteOption to Orphan.
+		modified := resourcemerge.BoolPtr(false)
+		existingOwners := existing.GetOwnerReferences()
+		resourcemerge.MergeOwnerRefs(modified, &existingOwners, required.GetOwnerReferences())
```
yue9944882: combining client-side patch and SSA looks odd to me. also if we're merging this, we should also deal with the merging of corresponding entries of `.metadata.managedFields`

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -271,12 +293,61 @@ func (m *ManifestWorkController) applyOneManifest(
...
 }
 
+func (m *ManifestWorkController) serverSideApply(
+	ctx context.Context,
+	gvr schema.GroupVersionResource,
+	required *unstructured.Unstructured,
+	config *workapiv1.ServerSideApplyConfig,
+	recorder events.Recorder) (*unstructured.Unstructured, bool, error) {
+	force := false
+	fieldManager := "work-agent"
```
yue9944882: nit: move it to public variables so that we can share the defaults from other places

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
qiujian16: In deleteOption feature, we need to ensure that the ownerrefs of appliedManifestwork should not exist. This is to deal with the case that user create a manifestwork and then update the manfiestwork's deleteOption to orphan. We need to get the existing owners on the resource, and ensure that owner of appliedManifestWork exists or not based on deleteOption. Since ownerrefs is a slice, I think we replace  the whole slice here.

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
qiujian16: create a PR to put it in api repo https://github.com/open-cluster-management-io/api/pull/164

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
deads2k: > I think this is needed when user changes the deleteOption to Orphan. We should remove the ownerref of ApplieManifestWork on this resource. Added a comment to explain this.

Good catch, I missed that case.  I'll re-review for it.

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
qiujian16: toclarify, this code is not to generate a merge patch, but update the required object by changing the ownerrefs on it.

qiujian16: I update this part. We should treat ownerrefs separately in server apply case, so it would not mess up with other field. Also add an integration test.

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
qiujian16: I update this part. We should treat ownerrefs separately in server apply case, so it would not mess up with other field. Also add an integration test.

yue9944882 (commented):
LGTM

PR #147, opened 2022-08-25
handle nil of ApplyResult.Result

elgnay:


qiujian16:
this won't be needed

/close

PR #149, opened 2022-08-29
move apply method to seperate package

zhujian7:


pkg/spoke/spoketesting/manifestwork_helpers.go
```diff
@@ -51,6 +51,12 @@ func NewSecret(name, namespace string, content string) *corev1.Secret {
 	}
 }
 
+func NewSecretWithType(name, namespace string, content string, t corev1.SecretType) *corev1.Secret {
```
qiujian16, line 54: where do we use this func?

zhujian7: in the update apply unit test: https://github.com/open-cluster-management-io/work/pull/149/files#diff-d09459436f4e418295b505389c79e8f8fdc029833ef7558dffef4209339841b5R345

PR #151, opened 2022-08-30
fix Missing filter in issues#150

TYTaO:
Use hubHash to distinguish resources like other controllers

### DIFF
--- a/pkg/spoke/controllers/finalizercontroller/appliedmanifestwork_finalize_controller.go
+++ b/pkg/spoke/controllers/finalizercontroller/appliedmanifestwork_finalize_controller.go
@@ -36,6 +36,7 @@ func NewAppliedManifestWorkFinalizeController(
 	spokeDynamicClient dynamic.Interface,
 	appliedManifestWorkClient workv1client.AppliedManifestWorkInterface,
 	appliedManifestWorkInformer workinformer.AppliedManifestWorkInformer,
+	hubHash string,
 ) factory.Controller {
 
 	controller := &AppliedManifestWorkFinalizeController{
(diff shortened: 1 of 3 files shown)


pkg/spoke/controllers/finalizercontroller/appliedmanifestwork_finalize_controller.go
```diff
@@ -48,6 +50,9 @@ func NewAppliedManifestWorkFinalizeController(
 	return factory.New().
 		WithInformersQueueKeyFunc(func(obj runtime.Object) string {
 			accessor, _ := meta.Accessor(obj)
+			if !strings.HasPrefix(accessor.GetName(), hubHash) {
+				return ""
```
qiujian16: we have a queueKeyFunc in helper that you can use. Also I think we should use WithFilteredInformersQueueKeyFunc

TYTaO: > 
Is the queueKeyFunc in helper this func？
https://github.com/open-cluster-management-io/work/blob/8f206974ddc0bb239e014768dbae19d0f37cad74/pkg/helper/helpers.go#L322
This func will return a manifestwork key, but in this controller we need a appliedmanifestwork key. May I can provide another func in helper?

qiujian16: ok, that will be great. And maybe also a filter func in helper that can be reused by multiple controllers.

TYTaO: > ok, that will be great. And maybe also a filter func in helper that can be reused by multiple controllers.

The changes have been committed

PR #152, opened 2022-09-05
check the executor subject permission for action apply

zhujian7:
issue reference: https://github.com/open-cluster-management-io/work/issues/153

### DIFF
--- a/deploy/spoke/clusterrole.yaml
+++ b/deploy/spoke/clusterrole.yaml
@@ -15,6 +15,7 @@ rules:
 - apiGroups: ["work.open-cluster-management.io"]
   resources: ["appliedmanifestworks/finalizers"]
   verbs: ["update"]
+# Allow agent to create subjectaccessreviews
 - apiGroups: ["authorization.k8s.io"]
   resources: ["subjectaccessreviews"]
   verbs: ["create"]
--- a/deploy/spoke/clusterrole.yaml
+++ b/deploy/spoke/clusterrole.yaml
@@ -15,4 +15,6 @@ rules:
 - apiGroups: ["work.open-cluster-management.io"]
   resources: ["appliedmanifestworks/finalizers"]
   verbs: ["update"]
-  
+- apiGroups: ["authorization.k8s.io"]
+  resources: ["subjectaccessreviews"]
+  verbs: ["create"]
--- a/pkg/spoke/auth/auth.go
+++ b/pkg/spoke/auth/auth.go
@@ -4,9 +4,9 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	authorizationv1 "k8s.io/api/authorization/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes"
@@ -31,6 +31,19 @@ type ExecutorValidator interface {
 	// if there is no permission will return a kubernetes forbidden error.
 	Validate(ctx context.Context, executor *workapiv1.ManifestWorkExecutor,
 		gvr schema.GroupVersionResource, namespace, name string, action ExecuteAction) error
+}
+
+type NotAllowedError struct {
+	Err         error
+	RequeueTime time.Duration
+}
+
+func (e *NotAllowedError) Error() string {
+	err := e.Err.Error()
+	if e.RequeueTime > 0 {
+		err = fmt.Sprintf("%s, will try again in %s", err, e.RequeueTime.String())
+	}
+	return err
 }
 
 func NewExecutorValidator(kubeClient kubernetes.Interface) ExecutorValidator {
@@ -61,7 +74,7 @@ func (v *sarValidator) Validate(ctx context.Context, executor *workapiv1.Manifes
 	var verbs []string
 	switch action {
 	case ApplyAction:
-		verbs = []string{"create", "update", "patch", "get", "list"}
+		verbs = []string{"create", "update", "patch", "get"}
 	case DeleteAction:
 		verbs = []string{"delete"}
 	default:
@@ -83,10 +96,11 @@ func (v *sarValidator) Validate(ctx context.Context, executor *workapiv1.Manifes
 	}
 
 	if !allowed {
-		return errors.NewForbidden(schema.GroupResource{
-			Group:    resource.Group,
-			Resource: resource.Resource,
-		}, resource.Name, fmt.Errorf("not allowed to %s the resource", strings.ToLower(string(action))))
+		return &NotAllowedError{
+			Err: fmt.Errorf("not allowed to %s the resource %s %s, name: %s",
+				strings.ToLower(string(action)), resource.Group, resource.Resource, resource.Name),
+			RequeueTime: 60 * time.Second,
+		}
 	}
 
 	return nil
--- a/pkg/spoke/auth/auth.go
+++ b/pkg/spoke/auth/auth.go
@@ -1,0 +1,139 @@
+package auth
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	authorizationv1 "k8s.io/api/authorization/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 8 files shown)


pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -154,6 +158,9 @@ func (m *ManifestWorkController) sync(ctx context.Context, controllerContext fac
 	errs := []error{}
 	// Apply resources on spoke cluster.
 	resourceResults := make([]applyResult, len(manifestWork.Spec.Workload.Manifests))
+	// check the Executor subject permission before applying
+	resourceResults = m.validateManifests(ctx, manifestWork.Spec.Executor,
```
qiujian16: shouldn't we validate each manifest separately?

pkg/spoke/helper/helper.go
```diff
@@ -0,0 +1,168 @@
...
+// DeleteAppliedResources deletes all given applied resources and returns those pending for finalization
+// If the uid recorded in resources is different from what we get by client, ignore the deletion.
+func DeleteAppliedResources(
+	ctx context.Context,
+	resources []workapiv1.AppliedManifestResourceMeta,
+	reason string,
+	dynamicClient dynamic.Interface,
+	recorder events.Recorder,
+	owner metav1.OwnerReference,
+	kubeClient kubernetes.Interface, executor *workapiv1.ManifestWorkExecutor) ([]workapiv1.AppliedManifestResourceMeta, []error) {
```
qiujian16: I think you want to input validator here, we need to consider cache in the future.

pkg/spoke/helper/helper.go
```diff
@@ -0,0 +1,168 @@
...
+
+		err = validator.ValidateDelete(ctx, executor, gvr, resource.Namespace, resource.Name)
+		if err != nil {
+			if !apierrors.IsForbidden(err) {
+				errs = append(errs, fmt.Errorf(
+					"failed to check delete permission for resource %v with key %s/%s: %w",
+					gvr, resource.Namespace, resource.Name, err))
+				continue
+			}
+			// TODO: consider to reflect the not allowed error on the manifest work condition
```
qiujian16: this will be certainly needed. I actually think all delete action should be put in the separate PR together with how we handle condition.

pkg/spoke/controllers/finalizercontroller/appliedmanifestwork_finalize_controller.go
```diff
@@ -96,12 +105,25 @@ func (m *AppliedManifestWorkFinalizeController) syncAppliedManifestWork(ctx cont
...
+		return err
+	}
+
 	// Work is deleting, we remove its related resources on spoke cluster
 	// We still need to run delete for every resource even with ownerref on it, since ownerref does not handle cluster
 	// scoped resource correctly.
 	reason := fmt.Sprintf("manifestwork %s is terminating", appliedManifestWork.Spec.ManifestWorkName)
-	resourcesPendingFinalization, errs := helper.DeleteAppliedResources(
-		ctx, appliedManifestWork.Status.AppliedResources, reason, m.spokeDynamicClient, controllerContext.Recorder(), *owner)
+	resourcesPendingFinalization, errs := spokehelper.DeleteAppliedResources(
```
qiujian16: we also need to update condition for work here, to declare the remaining resource and err

pkg/spoke/controllers/finalizercontroller/manifestwork_finalize_controller.go
```diff
@@ -68,14 +68,17 @@ func (m *ManifestWorkFinalizeController) sync(ctx context.Context, controllerCon
...
 		if err != nil {
 			return err
 		}
 	case err != nil:
 		return err
 	case !manifestWork.DeletionTimestamp.IsZero():
-		err := m.deleteAppliedManifestWork(ctx, appliedManifestWorkName)
+		// delete the applied manifestwork, but orphan the dependent resources, we will
+		// check the executor subject permission before deleting these resources
+		orphanDependents := manifestWork.Spec.Executor != nil
```
qiujian16: hrm If we orphan delete the appliedwork, will lose the ownership of the resources, right? it could be conflict with deleteOption.

zhujian7:
@qiujian16 @elgnay PTAL

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -187,18 +188,41 @@ func (m *ManifestWorkController) sync(ctx context.Context, controllerContext fac
 		manifestCondition.Conditions = append(manifestCondition.Conditions, buildAppliedStatusCondition(result))
 
 		newManifestConditions = append(newManifestConditions, manifestCondition)
+
+		if apierrors.IsForbidden(result.Error) {
+			// If it is a forbidden error, after the condition is constructed, we set the error to nil
```
qiujian16: I think we should let validator to return the requeue time in the error. We can construct a special error struct in validator. Here we get the requeue time for each manifest in the work, and find the min one. This will be usefule when we implement validate cache later. Because the validator should use cache ttl to calculate the requeue time.

pkg/spoke/auth/auth.go
```diff
@@ -33,6 +33,15 @@ type ExecutorValidator interface {
 		gvr schema.GroupVersionResource, namespace, name string, action ExecuteAction) error
 }
 
+type NotAllowedError struct {
+	Err         error
+	RequeueTime time.Duration
+}
+
+func (e *NotAllowedError) Error() string {
+	return e.Err.Error()
```
qiujian16: nit: would you also add requeue time in the error message?

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -189,12 +189,16 @@ func (m *ManifestWorkController) sync(ctx context.Context, controllerContext fac
 
 		newManifestConditions = append(newManifestConditions, manifestCondition)
 
-		if apierrors.IsForbidden(result.Error) {
-			// If it is a forbidden error, after the condition is constructed, we set the error to nil
-			// and requeue the item after a little while
-			needRequeue = true
+		// If it is a forbidden error, after the condition is constructed, we set the error to nil
+		// and requeue the item
+		var autoError = &auth.NotAllowedError{}
```
qiujian16: auto or notAllowedError?

qiujian16 (commented):
just two nit comment, others looks good

/approve

pkg/spoke/auth/auth.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	sa := executor.Subject.ServiceAccount
+	if sa == nil {
+		return fmt.Errorf("the executor service account is nil")
+	}
+
+	var verbs []string
+	switch action {
+	case ApplyAction:
```
qiujian16, line 76: list is not necessary, right?

deploy/spoke/clusterrole.yaml
```diff
@@ -15,4 +15,6 @@ rules:
 - apiGroups: ["work.open-cluster-management.io"]
   resources: ["appliedmanifestworks/finalizers"]
   verbs: ["update"]
-  
\ No newline at end of file
+- apiGroups: ["authorization.k8s.io"]
```
qiujian16, line 19: nit: you may want to add this in operator at first.

pkg/spoke/auth/auth.go
zhujian7: I think it is not necessary, but the [proposal](https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/34-work-executor-group#workflow) has the "list" verb, so do we need to update the proposal?

deploy/spoke/clusterrole.yaml
zhujian7: yes, I  added it https://github.com/open-cluster-management-io/registration-operator/pull/273

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
zhujian7: sorry, typo, changed to authError

pkg/spoke/auth/auth.go
zhujian7: added

pkg/spoke/auth/auth.go
qiujian16: yes, I think so.

Issue #156, opened 2022-09-06
add a validate webhook to check the `execute-as` permission on the hub cluster


zhujian7:
this is a sub-task of https://github.com/open-cluster-management-io/work/issues/124 , we need to setup a validating webhook admission controller to check whether the requesting hub users have sufficient permission to manipulate the work object, see https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/34-work-executor-group#admission-control-via-subjectaccessreview

zhujian7:
In our expectation, the webhook will check if the user manipulating the manifestwork has `execute-as` permission for the executor(`manifestwork.spec.executor`), if the executor is nil, check if it has `execute-as` permission for the default executor `klusterlet-work-sa`. And this will impact all manifestworks already exist now, so for backward compatibility, we are going to rollout the webhook step by step:

Release 0.8.0:
- [ ] implement the webhook, and make a feature gate "checking executor nil permission" **disabled** by default.
- [ ] announce that all users who are using manifestwork with nil executor are required to 
  - be granted with `execute-as` permission for the default executor `klusterlet-work-sa`  **OR**
  - use a new executor in their manifestwork

Release 0.9.0 or 0.10.0:
- [ ] make the feature gate "checking executor nil permission" **enabled** by default.

PR #158, opened 2022-09-13
validate manifest work executor permission on hub cluster

zhujian7:

Issue reference: https://github.com/open-cluster-management-io/work/issues/156


zhujian7:
cc @qiujian16 @elgnay

qiujian16:
it looks good in general, but it needs e2e test.

qiujian16:
also how a user can set it when installing klusterlet?

zhujian7:
> also how a user can set it when installing klusterlet?

The webhook will be deployed on the hub cluster, so this should be configured on the `ClusterManager`, right? I think we can use the [clustermanager.spec.registrationConfiguration.featureGates](https://github.com/open-cluster-management-io/api/blob/efae592d440845b83693239cedfefe4069508469/operator/v1/types.go#L59) to do that.

zhujian7:
/hold
need to add e2e tests

pkg/webhook/manifestwork_webhook.go
```diff
@@ -129,3 +152,57 @@ func (a *ManifestWorkAdmissionHook) validateManifest(manifest []byte) error {
...
+		}
+	}
+
+	if work.Spec.Executor == nil {
+		executor = &workv1.ManifestWorkExecutor{
+			Subject: workv1.ManifestWorkExecutorSubject{
+				Type: workv1.ExecutorSubjectTypeServiceAccount,
+				ServiceAccount: &workv1.ManifestWorkSubjectServiceAccount{
+					// give the default value "system:serviceaccount::klusterlet-work-sa"
+					Namespace: "",                   // TODO: Not sure what value is reasonable
```
zhujian7, line 183: we use empty here, which means hub users who want to use nil executor need `execute-as` permission on manifestwork with resource name `system:serviceaccount::klusterlet-work-sa`

zhujian7:
/hold cancel
e2e tests are added
@qiujian16 ptal

qiujian16:
/approve

need another review from @elgnay

PR #159, opened 2022-09-20
check escalation for executor subject

### LINKED ISSUE #153: check the executor subject permission for action apply
this is a sub-task of https://github.com/open-cluster-management-io/work/issues/124 , we need to check the executor subject permission for action apply

proposal link: https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/34-work-executor-group

zhujian7:
To fix https://github.com/open-cluster-management-io/work/issues/153#issuecomment-1248855226

### DIFF
--- a/pkg/spoke/auth/auth.go
+++ b/pkg/spoke/auth/auth.go
@@ -7,9 +7,14 @@ import (
 	"time"
 
 	authorizationv1 "k8s.io/api/authorization/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
 
 	workapiv1 "open-cluster-management.io/api/work/v1"
 )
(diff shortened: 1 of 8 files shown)


test/integration/executor_test.go
```diff
@@ -209,4 +209,160 @@ var _ = ginkgo.Describe("ManifestWork Executor Subject", func() {
...
+			roleName := "role1"
+			_, err = spokeKubeClient.RbacV1().ClusterRoles().Create(
+				context.TODO(), &rbacv1.ClusterRole{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: roleName,
+					},
+					Rules: []rbacv1.PolicyRule{
+						{
+							// with "escalate" verb
+							// Verbs: []string{"create", "update", "patch", "get", "list", "delete", "escalate"},
```
zhujian7, line 313: it is weird that `Verbs: []string{"create", "update", "patch", "get", "list", "delete", "escalate"},` does not work, but `Verbs: []string{"*"},` does

zhujian7:
cc @qiujian16 @elgnay

pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -84,7 +86,7 @@ func NewManifestWorkController(
 		hubHash:                   hubHash,
 		restMapper:                restMapper,
 		appliers:                  apply.NewAppliers(spokeDynamicClient, spokeKubeClient, spokeAPIExtensionClient),
-		validator:                 auth.NewExecutorValidator(spokeKubeClient),
+		validator:                 auth.NewExecutorValidator(spokeKubeConfig, spokeKubeClient),
```
qiujian16: I think you may want to init this outside of the controller, you can reuse it when delete later.

pkg/spoke/auth/auth.go
```diff
@@ -97,15 +117,61 @@ func (v *sarValidator) Validate(ctx context.Context, executor *workapiv1.Manifes
...
+	dynamicClient, err := v.newImpersonateClientFunc(v.config, username(sa.Namespace, sa.Name))
+	if err != nil {
+		return err
+	}
+
+	_, err = dynamicClient.Resource(gvr).Namespace(namespace).Create(ctx, obj, metav1.CreateOptions{
+		DryRun: []string{"All"},
+	})
+	if apierrors.IsAlreadyExists(err) {
+		return nil
```
elgnay: If the k8s resource exists and its spec is different from the required one, we may need to update the k8s resource and see if the user has the permission or not. What do you think?

zhujian7: I checked even if the k8s resource exists, the API server will return the forbidden error when receiving an escalated request.
And I added 2 integration tests to cover these.

Issue #154, opened 2022-09-06
check the executor subject permission for action delete


zhujian7:
this is a sub-task of https://github.com/open-cluster-management-io/work/issues/124 , we need to check the executor subject permission for action delete

proposal link: https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/34-work-executor-group

zhujian7:
PR was merged
/close

PR #160, opened 2022-10-08
check the executor subject permission for delete

zhujian7:
issue reference: https://github.com/open-cluster-management-io/work/issues/154

### DIFF
--- a/pkg/spoke/auth/auth.go
+++ b/pkg/spoke/auth/auth.go
@@ -3,7 +3,6 @@ package auth
 import (
 	"context"
 	"fmt"
-	"strings"
 	"time"
 
 	authorizationv1 "k8s.io/api/authorization/v1"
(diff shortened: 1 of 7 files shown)


pkg/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -274,15 +274,18 @@ func (m *ManifestWorkController) applyOneManifest(
 		return result
 	}
 
+	// check if the to be applied resource should be owned by the manifest work
```
qiujian16: resource to be applied

pkg/spoke/auth/auth.go
```diff
@@ -91,14 +81,9 @@ func (v *sarValidator) Validate(ctx context.Context, executor *workapiv1.Manifes
...
-	var verbs []string
-	switch action {
-	case ApplyAction:
-		verbs = []string{"create", "update", "patch", "get"}
-	case DeleteAction:
-		verbs = []string{"delete"}
-	default:
-		return fmt.Errorf("execute action %s is invalid", action)
+	verbs := []string{"create", "update", "patch", "get"}
+	if ownedByTheWork {
```
qiujian16, line 85: we should add a TODO here indicating the potential risk.

zhujian7: added some comments, ptal

PR #161, opened 2022-10-19
Delete unmanaged AppliedManifestWork

qiujian16:

1. record agent id in appliedmanifestwork
2. delete the appliedmanifestwork if it is owned by this agent but lose the management.

### DIFF
--- a/pkg/spoke/controllers/finalizercontroller/appliedmanifestwork_finalize_controller.go
+++ b/pkg/spoke/controllers/finalizercontroller/appliedmanifestwork_finalize_controller.go
@@ -57,7 +57,7 @@ func NewAppliedManifestWorkFinalizeController(
 
 func (m *AppliedManifestWorkFinalizeController) sync(ctx context.Context, controllerContext factory.SyncContext) error {
 	appliedManifestWorkName := controllerContext.QueueKey()
-	klog.V(4).Infof("Reconciling ManifestWork %q", appliedManifestWorkName)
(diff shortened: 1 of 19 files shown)


qiujian16:
/hold

still need tests

pkg/spoke/controllers/finalizercontroller/appliedmanifestwork_finalize_controller.go
```diff
@@ -50,7 +51,7 @@ func NewAppliedManifestWorkFinalizeController(
 		WithFilteredEventsInformersQueueKeyFunc(func(obj runtime.Object) string {
 			accessor, _ := meta.Accessor(obj)
 			return accessor.GetName()
-		}, helper.AppliedManifestworkHubHashFilter(hubHash), appliedManifestWorkInformer.Informer()).
+		}, helper.AppliedManifestworkAgentIDFilter(agentID), appliedManifestWorkInformer.Informer()).
```
zhujian7, line 54: We don't need the `hubHash` filter anymore because the appliedmanifestwork_finalize_controller needs to handle the unmanaged_appliedmanifestwork which is in deleting status, right?

qiujian16: it is still needed for appliedmanifestwork_controller.go since it should only handle those with correct hash.

elgnay:
Looks good to me. The only concern is that any CRD change on managed cluster may result in CRD conflict in the multiple hub scenario if the two hubs have different version.

PR #164, opened 2022-11-02
use controller runtime to implement webhook

ldpliu:
Signed-off-by: ldpliu <daliu@redhat.com>
https://github.com/open-cluster-management-io/OCM/issues/81


pkg/webhook/v1/webhook.go
```diff
@@ -0,0 +1,27 @@
+package v1
+
+import (
+	"k8s.io/client-go/kubernetes"
+	v1 "open-cluster-management.io/api/work/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+type ManifestWorkWebhook struct {
+	kubeClient kubernetes.Interface
```
qiujian16, line 10: I do not thinks it is needed in this webhook, right?

ldpliu: It is used here: https://github.com/open-cluster-management-io/work/pull/164/files#diff-c05eb54e7b9381bee7827595cf0c481f6c38395bdafaccfb692136bfddba67aaR158

qiujian16:
@elgnay PTAL

Issue #155, opened 2022-09-06
cache the subject access review results


zhujian7:
A sub-task of https://github.com/open-cluster-management-io/work/issues/124 , there must be a lot of subject access review requests to the API server per the [proposal](https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/34-work-executor-group), we need to find a way to cache the subject access review results.

PR #165, opened 2022-11-04
cache the executor validation results

zhujian7:

issue reference: https://github.com/open-cluster-management-io/work/issues/155


pkg/spoke/controllers/cachecontroller/executor_cache_controller.go
```diff
@@ -0,0 +1,323 @@
+package cachecontroller
+
+import (
+	"context"
+	"fmt"
```
qiujian16: I think it is better to put it in the auth/cache package.  I think it is better to have a cache.Start to start the routine of such controller

pkg/spoke/controllers/cachecontroller/indexes.go
```diff
@@ -0,0 +1,46 @@
...
+
+	rbacv1 "k8s.io/api/rbac/v1"
+)
+
+const (
+	byClusterRole = "byClusterRole"
+	byRole        = "byRole"
+)
+
+func crbIndexByClusterRole(obj interface{}) ([]string, error) {
```
qiujian16: indexByClusterRoleRef

pkg/spoke/controllers/cachecontroller/executor_cache_controller.go
```diff
@@ -0,0 +1,318 @@
...
+		manifestWorkLister: manifestWorkLister,
+		restMapper:         restMapper,
+		executorCaches:     executorCaches,
+		sarCheckerFn:       sarCheckerFn,
+	}
+
+	return factory.New().
+		WithInformersQueueKeysFunc(
+			roleEnqueueFu(rbInformer.Informer().GetIndexer(), executorCaches),
+			rInformer.Informer()).
```
qiujian16: we might need to use specific AddEventHandler with bare informer here. I think it is easy to handle add/update event, but how about delete event?

pkg/spoke/auth/cache_auth.go
```diff
@@ -0,0 +1,114 @@
...
+		Version:       gvr.Version,
+		ExecuteAction: GetExecuteAction(ownedByTheWork),
+	}
+
+	if allow, ok := v.executorCaches.Get(executorKey, dimension); !ok {
+		err := v.validator.checkSubjectAccessReviews(ctx, sa, gvr, namespace, name, ownedByTheWork)
+		UpdateSARCheckResultToCache(v.executorCaches, executorKey, dimension, err)
+		if err != nil {
+			return err
+		}
```
qiujian16: move return v.validator.checkEscalation(ctx, sa, gvr, namespace, name, obj) to here just to make it clearer that it should only be called when there is not cache.

pkg/spoke/auth/cache_store.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func (c *ExecutorCaches) RemoveDimensionCaches(executor string) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	delete(c.items, executor)
+}
+
+func (c *ExecutorCaches) GetCacheItems() map[string]*DimensionCaches {
+	c.lock.Lock()
```
qiujian16: hrm, why we need this? can  a func get this and mutate items out of lock?
or you want a copy of the items to be returned?

pkg/spoke/auth/cache_store.go
```diff
@@ -0,0 +1,183 @@
...
+	defer c.lock.Unlock()
+	dimensionCaches := &DimensionCaches{
+		lock:  sync.RWMutex{},
+		items: make(map[string]CacheValue),
+	}
+	c.items[executor] = dimensionCaches
+}
+
+func (c *ExecutorCaches) GetDimensionCaches(executor string) (*DimensionCaches, bool) {
```
qiujian16: i am not quite sure about returning point here,  is it safe?

pkg/spoke/auth/cache_store.go
```diff
@@ -0,0 +1,183 @@
...
+
+import (
+	"crypto/sha256"
+	"encoding/json"
+	"fmt"
+	"sync"
+
+	"k8s.io/klog/v2"
+)
+
```
qiujian16: I am not so sure about the cache struct. some doc might be needed.

pkg/spoke/auth/cache_auth.go
zhujian7: Whether there is a cache or not, we need to call the checkEscalation. because whether it will escalate not only depends on the role/rolebinding but also on the content of the resource to be applied, which may change

pkg/spoke/auth/cache_store.go
zhujian7: I want a copy of the items to be returned. so in the cache controller we can use a for-range loop to update these caches.

zhujian7: This has been fixed

pkg/spoke/auth/cache_store.go
zhujian7: changed it to an internal method

zhujian7:
@qiujian16 @elgnay PTAL

pkg/spoke/spokeagent.go
```diff
@@ -57,6 +57,7 @@ func (o *WorkloadAgentOptions) AddFlags(cmd *cobra.Command) {
 	flags.Float32Var(&o.QPS, "spoke-kube-api-qps", o.QPS, "QPS to use while talking with apiserver on spoke cluster.")
 	flags.IntVar(&o.Burst, "spoke-kube-api-burst", o.Burst, "Burst to use while talking with apiserver on spoke cluster.")
 	flags.DurationVar(&o.StatusSyncInterval, "status-sync-interval", o.StatusSyncInterval, "Interval to sync resource status to hub.")
+	flags.BoolVar(&o.EnableExecutorCaches, "enable-executor-caches", o.EnableExecutorCaches, "Whether to enable the executor caches function.")
```
zhujian7: Should we set this as a feature gate? so users can use klusterlet to config.
@qiujian16 WDYT?

pkg/spoke/auth/cache/auth.go
```diff
@@ -0,0 +1,152 @@
...
+// It's an error to call Start more than once.
+// Start blocks; call via go.
+func (v *sarCacheValidator) Start(ctx context.Context) {
+	// initialize the caches skelton in order to let others caches operands know which caches are necessary,
+	// otherwise, the roleBindingExecutorsMapper and clusterRoleBindingExecutorsMapper in the cache controller
+	// have no chance to initialize after the work pod restarts
+	v.manifestWorkExecutorCachesLoader.loadAllValuableCaches(v.executorCaches)
+
+	v.spokeInformer.Start(ctx.Done())
+	v.spokeInformer.WaitForCacheSync(ctx.Done())
```
qiujian16: should this be done already in the controller code when start?

pkg/spoke/auth/cache/auth.go
```diff
@@ -0,0 +1,152 @@
...
+
+// updateSARCheckResultToCache updates the subjectAccessReview checking result to the executor cache
+func updateSARCheckResultToCache(executorCaches *store.ExecutorCaches, executorKey string,
+	dimension store.Dimension, result error) {
+	t, f := true, false
+	if result == nil {
+		executorCaches.Upsert(executorKey, dimension, &t)
+	}
+
+	var authError = &basic.NotAllowedError{}
```
qiujian16: nit no need for this line, this could reduce one variable

pkg/spoke/auth/cache/auth.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+	}
+
+	return v.validator.CheckEscalation(ctx, sa, gvr, namespace, name, obj)
+}
+
+// updateSARCheckResultToCache updates the subjectAccessReview checking result to the executor cache
+func updateSARCheckResultToCache(executorCaches *store.ExecutorCaches, executorKey string,
+	dimension store.Dimension, result error) {
+	t, f := true, false
```
qiujian16: using pointers library?

pkg/spoke/auth/cache/auth.go
```diff
@@ -0,0 +1,152 @@
...
+	dimension := store.Dimension{
+		Namespace:     namespace,
+		Name:          name,
+		Resource:      gvr.Resource,
+		Group:         gvr.Group,
+		Version:       gvr.Version,
+		ExecuteAction: store.GetExecuteAction(ownedByTheWork),
+	}
+
+	allowed, ok := v.executorCaches.Get(executorKey, dimension)
```
qiujian16: nit seems no need to return two here? allows==nil means not found

pkg/spoke/auth/cache/auth_test.go
```diff
@@ -0,0 +1,243 @@
...
+		},
+		"sa nil": {
+			executor: &workapiv1.ManifestWorkExecutor{
+				Subject: workapiv1.ManifestWorkExecutorSubject{
+					Type: workapiv1.ExecutorSubjectTypeServiceAccount,
+				},
+			},
+			expect: fmt.Errorf("the executor service account is nil"),
+		},
+		"forbideen": {
```
qiujian16: forbidden

pkg/spoke/auth/cache/auth_test.go
```diff
@@ -0,0 +1,243 @@
...
+		},
+	}
+
+	tests := map[string]struct {
+		executor  *workapiv1.ManifestWorkExecutor
+		namespace string
+		name      string
+		expect    error
+	}{
+		"forbideen": {
```
qiujian16: forbidden

pkg/spoke/auth/cache/binding_event_handler.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+}
+
+func (h *roleBindingEventHandler) OnUpdate(oldObj, newObj interface{}) {
+	h.OnAdd(newObj)
+}
+
+func (h *roleBindingEventHandler) OnDelete(obj interface{}) {
+
+	if rb, ok := obj.(*rbacapiv1.RoleBinding); ok {
```
qiujian16, line 30: add some comment here

pkg/spoke/auth/cache/executor_cache_controller.go
```diff
@@ -0,0 +1,314 @@
...
+
+	crbInformer.Informer().AddEventHandler(&clusterRoleBindingEventHandler{
+		enqueueUpsertFunc: controller.bindingResourceUpsertEnqueueFn(syncCtx),
+		enqueueDeleteFunc: controller.bindingResourceDeleteEnqueueFn(syncCtx),
+	})
+
+	return factory.New().
+		WithSyncContext(syncCtx).
+		WithInformersQueueKeysFunc(
+			roleEnqueueFu(rbInformer.Informer().GetIndexer(), controller.executorCaches),
```
qiujian16: curious, why not put enqueufunc and indexers in controller?

pkg/spoke/auth/cache/executor_cache_controller.go
```diff
@@ -0,0 +1,314 @@
...
+	ResyncInterval = 10 * time.Minute
+)
+
+// CacheController is to reconcile the executor auth result for manfiestwork workloads
+// on spoke cluster.
+type CacheController struct {
+	executorCaches                   *store.ExecutorCaches
+	sarCheckerFn                     SubjectAccessReviewCheckFn
+	manifestWorkExecutorCachesLoader manifestWorkExecutorCachesLoader
+	bindingExecutorsMapper           *safeMap
```
qiujian16: add a comment here. what is the purpose for this map

pkg/spoke/auth/cache/executor_cache_controller.go
```diff
@@ -0,0 +1,314 @@
...
+		klog.V(4).Infof("Update executor cache for executorKey: %s, dimension: %+v result: %v",
+			executorKey, v.Dimension, err)
+		updateSARCheckResultToCache(c.executorCaches, executorKey, v.Dimension, err)
+		return nil
+	}
+}
+
+func (c *CacheController) cleanupUnnecessaryCache() {
+	retainableCache := store.NewExecutorCache()
+	c.manifestWorkExecutorCachesLoader.loadAllValuableCaches(retainableCache)
```
qiujian16, line 296: comment here

pkg/spoke/spokeagent.go
qiujian16: yes, I think we should do that.

zhujian7:
@qiujian16 @elgnay PTAL

qiujian16 (commented):
/approve

PR #172, opened 2022-12-27
delete old applied work after its new work is applied

skeeey:
Signed-off-by: Wei Liu <liuweixa@redhat.com>

### DIFF
--- a/pkg/spoke/controllers/finalizercontroller/unmanaged_appliedmanifestwork_controller.go
+++ b/pkg/spoke/controllers/finalizercontroller/unmanaged_appliedmanifestwork_controller.go
@@ -23,8 +23,9 @@ import (
 
 const byWorkNameAndAgentID = "UnManagedAppliedManifestWork-byWorkNameAndAgentID"
 
-// ManifestWorkFinalizeController handles cleanup of manifestwork resources before deletion is allowed.
+// UnManagedAppliedWorkController deletes unmanaged applied works.
 type UnManagedAppliedWorkController struct {
+	manifestWorkLister         worklister.ManifestWorkNamespaceLister
 	appliedManifestWorkClient  workv1client.AppliedManifestWorkInterface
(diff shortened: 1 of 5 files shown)


qiujian16 (commented):
/approve
/lgtm

pkg/spoke/controllers/finalizercontroller/unmanaged_appliedmanifestwork_controller.go
```diff
@@ -71,6 +79,26 @@ func (m *UnManagedAppliedWorkController) sync(ctx context.Context, controllerCon
 		// work not found, could have been deleted, do nothing.
 		return nil
 	}
+	if err != nil {
+		return err
+	}
+
+	// We delete the old applied work until the work of this applied work is applied.
```
BenamarMk, line 86: Do you mean `We don't delete` instead of `We delete`. Let me rephrase it to see if I can understand it:
```
We don't delete the old AppliedManifestWork until the new one is applied
```
Also, where is the delete???

elgnay: > Do you mean `We don't delete` instead of `We delete`. Let me rephrase it to see if I can understand it:
> 
> ```
> We don't delete the old AppliedManifestWork until the new one is applied
> ```

Yes, you are right. There is a typo in the comment. It should be
```
We don't delete the old AppliedManifestWork until the new one is applied
```

> 
> Also, where is the delete???

The delete logic can be found here:
https://github.com/open-cluster-management-io/work/blob/main/pkg/spoke/controllers/finalizercontroller/unmanaged_appliedmanifestwork_controller.go#L109-L115

skeeey: yes, this pr postpones the deletion operation,  we need wait the new appliedmanifest applied to avoid deleting the old applied work prematurely

PR #174, opened 2023-01-04
clear the comments for deleting unmanaged appliedmanifestworks

skeeey:
Signed-off-by: Wei Liu <liuweixa@redhat.com>

### DIFF
--- a/pkg/spoke/controllers/finalizercontroller/unmanaged_appliedmanifestwork_controller.go
+++ b/pkg/spoke/controllers/finalizercontroller/unmanaged_appliedmanifestwork_controller.go
@@ -83,10 +83,9 @@ func (m *UnManagedAppliedWorkController) sync(ctx context.Context, controllerCon
 		return err
 	}
 
-	// We delete the old applied work until the work of this applied work is applied.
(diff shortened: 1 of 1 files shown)


pkg/spoke/controllers/finalizercontroller/unmanaged_appliedmanifestwork_controller.go
```diff
@@ -83,9 +83,9 @@ func (m *UnManagedAppliedWorkController) sync(ctx context.Context, controllerCon
 		return err
 	}
 
-	// We delete the old applied work until the work of this applied work is applied.
-	// This will avoid to delete the old applied work prematurely, if an applied work is an
-	// owner of its applied resource, if we delete the old applied work prematurely, this will
+	// We delete the old appliedmanifestwork only when the work of its new appliedmanifestwork is applied.
+	// This will avoid to delete the old appliedmanifestwork prematurely, if an appliedmanifestwork is an
```
qiujian16: We delete the old AppliedManifestWork only when the related ManifestWork is applied with the new AppliedManifestWork as the new owner. This can avoid deleting the old AppliedManifestWork prematurely before the new AppliedManifestWork takes the ownership of the applied resources.

PR #175, opened 2023-01-10
Add informer setup for placemanifestwork

qiujian16:
- add index and informer to watch resources
- build binary for hub manager


serngawy:
I guess its missing adding the pkg/cmd/hub to the main at the cmd/work/main ? OR may be better have another main for placeManifestwork controller under cmd/hub/main ?

qiujian16:
added. Thanks for reminding :)

serngawy (approved):

PR #176, opened 2023-01-13
Split logic of controller to multiple reconcilers

qiujian16:
This could make code clearer
also update the crd fix


qiujian16:
/assign @serngawy 

I have not add tests yet, this just gives an idea on how reconcile work

serngawy (approved):
I will approve this PR to be able to start work on the manifestwork creation and the unit test can be added later in another PR

PR #177, opened 2023-01-18
Add placeManifestWork create,update and delete 

serngawy:
Signed-off-by: melserngawy <melserng@redhat.com>


pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
```diff
@@ -18,5 +27,92 @@ type deployReconciler struct {
 
 func (d *deployReconciler) reconcile(ctx context.Context, pw *workapiv1alpha1.PlaceManifestWork) (*workapiv1alpha1.PlaceManifestWork, reconcileState, error) {
 	// TODO the main manifestwork create/update/delete logic, we may also need to manage the status condition for whether manifestwork is created here
+	klog.Info("reconciling placeManifestWork ", pw.Name)
+
+	placement, err := d.placementLister.Placements(pw.Namespace).Get(pw.Spec.PlacementRef.Name)
+	if err != nil {
```
qiujian16, line 32: should treat NotFoundError differently. So we set condition if it is not found and return reconcileStop and nil error, and return error directly with reconcileStop if it is not NotFoundError. e.g.

```go
switch {
case errors.IsNotFoundError:
    // set condition
   return pw, reconcileStop, nil
case error != nil:
   return pw, reconcileStop, err
}
```

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
```diff
@@ -18,5 +27,92 @@ type deployReconciler struct {
...
+		//klog.Info("my reconciling placement err ", err)
+		apimeta.SetStatusCondition(&pw.Status.Conditions, GetPlacementDecisionVerified(workapiv1alpha1.PlacementDecisionNotFound, ""))
+
+		return pw, reconcileContinue, fmt.Errorf("Failed get placement %w", err)
+	}
+
+	//klog.Info("my reconciling placement ", placement)
+	// TODO: Could be better to change the pw status to include list of selected clusters otherwise we need to list all the manifestworks belong to pw to figure which cluster Added or deleted
```
qiujian16: I think get manifestwork by lister is good. We cannot record clusters in status, since it could be many clusters and change frequently. Get by lister won't produce api call so it is appropriate here.

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
```diff
@@ -18,5 +27,92 @@ type deployReconciler struct {
...
+	req, _ := labels.NewRequirement(PlaceManifestWorkControllerNameLabelKey, selection.In, []string{pw.Name})
+	selector := labels.NewSelector().Add(*req)
+
+	manifestWorks, _ := d.manifestWorkLister.List(selector)
```
qiujian16: error handling is needed, just

```go
if err != nil {
    return pw, reconcileStop, err
}
```

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
```diff
@@ -18,5 +27,92 @@ type deployReconciler struct {
...
+	var existingClusters sets.String
+	for _, mw := range manifestWorks {
+		existingClusters.Insert(mw.Namespace)
+	}
+
+	added, deleted, _ := helper.GetClusters(d.placeDecisionLister, placement, existingClusters)
```
qiujian16: ditto: error handling

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
```diff
@@ -18,5 +27,92 @@ type deployReconciler struct {
...
+		apimeta.SetStatusCondition(&pw.Status.Conditions, GetPlacementDecisionVerified(workapiv1alpha1.PlacementDecisionEmpty, ""))
+		apimeta.SetStatusCondition(&pw.Status.Conditions, GetManifestworkApplied(workapiv1alpha1.PlacementDecisionEmpty, ""))
+	}
+
+	if len(deleted) > 0 {
+		// TODO Delete manifestwork
+	}
+
+	for cluster := range added {
+		mw, err := CreateManifestWork(pw, cluster)
```
qiujian16: this should be applyManifestWork. We would need to treat manifestwork update. you can use this https://github.com/open-cluster-management-io/api/blob/main/utils/work/v1/workapplier/workapplier.go

we can init the the applier during init the reconcile.

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
```diff
@@ -18,5 +27,92 @@ type deployReconciler struct {
...
+		if err != nil {
+			return pw, reconcileContinue, nil
+		}
+		_, err = d.workClient.WorkV1().ManifestWorks(cluster).Create(ctx, mw, metav1.CreateOptions{})
+		if err != nil {
+			return pw, reconcileContinue, nil
+		}
+	}
+
```
qiujian16, line 76: TODO update applied status here.

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
```diff
@@ -18,5 +27,92 @@ type deployReconciler struct {
...
+	if len(deleted) > 0 {
+		// TODO Delete manifestwork
+	}
+
+	for cluster := range added {
+		mw, err := CreateManifestWork(pw, cluster)
+		if err != nil {
+			return pw, reconcileContinue, nil
```
qiujian16: I would build a var errs []error, and append err here with continue. 

the errs can be returned finally with an aggregate err.

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
serngawy: done

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
serngawy: done

serngawy (commented):
Still work in progress (almost done). Waiting for approving the api [PR](https://github.com/open-cluster-management-io/api/pull/200)  and update the status.

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
```diff
@@ -18,5 +30,133 @@ type deployReconciler struct {
...
+		existingClusters.Insert(mw.Namespace)
+	}
+
+	addedClusters, deletedClusters, err := helper.GetClusters(d.placeDecisionLister, placement, existingClusters)
+	if err != nil {
+		return pw, reconcileContinue, utilerrors.NewAggregate(errs)
+	}
+	//klog.Info("Selected cluters count ", len(added), len(deleted))
+
+	workApplier := workapplier.NewWorkApplierWithTypedClient(d.workClient, d.manifestWorkLister)
```
qiujian16: this has the local cache. so you would want to init it at the very beginning when construct this structure. Putting applier as a member of the struct

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
```diff
@@ -18,5 +30,133 @@ type deployReconciler struct {
...
+			apimeta.SetStatusCondition(&pw.Status.Conditions, GetPlacementDecisionVerified(workapiv1alpha1.PlacementDecisionNotFound, ""))
+		}
+		return pw, reconcileContinue, fmt.Errorf("Failed get placement %w", err)
+	}
+
+	//klog.Info("my reconciling placement ", placement)
+	errs := []error{}
+	req, err := labels.NewRequirement(PlaceManifestWorkControllerNameLabelKey, selection.In, []string{pw.Name})
+	if err != nil {
+		errs = append(errs, err)
```
qiujian16: I would think directly return err here

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
```diff
@@ -18,5 +30,133 @@ type deployReconciler struct {
...
+	}
+
+	selector := labels.NewSelector().Add(*req)
+	manifestWorks, err := d.manifestWorkLister.List(selector)
+	if err != nil {
+		errs = append(errs, err)
```
qiujian16: same, directly return err, you would not want the following code to proceed if list here returns err.

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
```diff
@@ -18,5 +30,133 @@ type deployReconciler struct {
...
+		}
+	}
+
+	// Update created manifestWorks in case there are changes at ManifestWork or PlaceManifestWork
+	for _, mw := range manifestWorks {
+		if deletedClusters.Has(mw.Namespace) {
+			continue
+		}
+
+		// TODO: waiting for the PR https://github.com/open-cluster-management-io/api/pull/200 to be merged
```
qiujian16: I do not think you need this.

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
```diff
@@ -18,5 +30,133 @@ type deployReconciler struct {
...
+
+	// Delete manifestWork for deleted clusters
+	for cls := range deletedClusters {
+		err = workApplier.Delete(ctx, cls, pw.Name)
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	// Create manifestWork for added clusters
```
qiujian16, line 63: hrm I would think you only need to check if the manifestwork needs to be deleted by comparing it with deleted clusters. If not just call apply, since it is either create/update and applier will handle the rest pieces.

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
serngawy: there is a case if the ManifestWork changed and not matching the PlaceManifestWork. I guess we should re-enforce the ManifestWork defined in the PlaceManifestWork

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
serngawy: done

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
serngawy: okay

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
serngawy: okay

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
qiujian16: yeah, but applier should update those, right? It is not necessary to treat it differently with create process.

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
```diff
@@ -2,21 +2,170 @@ package placemanifestworkcontroller
...
+	// Delete manifestWork for deleted clusters
+	for cls := range deletedClusters {
+		err = d.workApplier.Delete(ctx, cls, pw.Name)
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	// Create manifestWork for added clusters
+	for cls := range addedClusters {
```
qiujian16, line 64: how about apply on all existing and added clusters then you would not need line 91~103

serngawy: okay, its just the existingClusters also contain the deleted clusters, that is reason for deleting first. Will make the loop on the existingCluster and then apply with the pmw.spec.mw assigned to the mw to be sure that if the mw.spec had changed we still enforce the pwm.spec.mw

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
jnpacker: This was the part I struggled with in the original implementation on the most efficient way to traverse with a large number of clusters. Worst case (2300).

I was going to suggest, you start with the existing loop from Ln 93, and instead of `continue` when its in deleted, actually delete the manifest work (since you said the existing Clusters includes those that need to be deleted).

Then just Add any new ones after like you already have the loop for.

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_status_reconcile.go
```diff
@@ -2,20 +2,88 @@ package placemanifestworkcontroller
...
 )
 
 // statusReconciler is to update placeManifestwork status.
 type statusReconciler struct {
-	workClient          workclientset.Interface
 	manifestWorkLister  worklisterv1.ManifestWorkLister
 	placeDecisionLister clusterlister.PlacementDecisionLister
 }
 
 func (d *statusReconciler) reconcile(ctx context.Context, pw *workapiv1alpha1.PlaceManifestWork) (*workapiv1alpha1.PlaceManifestWork, reconcileState, error) {
```
jnpacker, line 19: Is this reconciling on every ManifestWork.Status update?

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
serngawy: okay, make sense.

pkg/hub/controllers/placemanifestworkcontroller/placemanifestwork_deploy_reconcile.go
```diff
@@ -2,21 +2,164 @@ package placemanifestworkcontroller
...
 	"context"
+	"fmt"
+	"k8s.io/apimachinery/pkg/api/errors"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
+	//"k8s.io/klog/v2"
```
qiujian16: nit: remove this line

qiujian16 (commented):
looks good in general, would you add some unit test for this controller. We can add integration test /  e2e as the followup but unit test is necessary.

serngawy:
> looks good in general, would you add some unit test for this controller. We can add integration test / e2e as the followup but unit test is necessary.

sure, will add the unit test today. was just waiting for code logic review.

PR #180, opened 2023-02-02
Add initial e2e/integration for placeManifestWork

qiujian16:


serngawy:
Sounds good, it needs re-base AND as best practice rename/move files better to be in separate commit cause having all in same commit we may lose the history of renamed/moved files.

serngawy (approved):

PR #182, opened 2023-02-07
Add validate webhook

serngawy:
Signed-off-by: melserngawy <melserng@redhat.com>

### DIFF
--- a/pkg/webhook/v1/manifestwork_validating.go
+++ b/pkg/webhook/v1/manifestwork_validating.go
@@ -59,7 +59,7 @@ func (r *ManifestWorkWebhook) validateRequest(newWork, oldWork *workv1.ManifestW
 		return apierrors.NewBadRequest("manifests should not be empty")
 	}
 
-	if err := validateManifests(newWork); err != nil {
+	if err := ValidateManifests(newWork.Spec.Workload.Manifests); err != nil {
 		return apierrors.NewBadRequest(err.Error())
 	}
 
@@ -75,9 +75,13 @@ func (r *ManifestWorkWebhook) validateRequest(newWork, oldWork *workv1.ManifestW
 	return validateExecutor(r.kubeClient, newWork, req.UserInfo)
 }
 
-func validateManifests(work *workv1.ManifestWork) error {
+func ValidateManifests(manifests []workv1.Manifest) error {
+	if len(manifests) == 0 {
+		return apierrors.NewBadRequest("Workload manifests should not be empty")
+	}
+
 	totalSize := 0
-	for _, manifest := range work.Spec.Workload.Manifests {
+	for _, manifest := range manifests {
 		totalSize = totalSize + manifest.Size()
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/webhook/v1alpha1/placemanifestwork_validating.go
```diff
@@ -55,6 +55,12 @@ func (r *PlaceManifestWorkWebhook) validateRequest(newPlaceWork *workv1alpha1.Pl
 }
 
 func validatePlaceManifests(placeWork *workv1alpha1.PlaceManifestWork) error {
+	if placeWork.Spec.PlacementRef.Name == "" {
```
qiujian16: this should be validated by openapi schema here https://github.com/open-cluster-management-io/api/blob/main/work/v1alpha1/types.go#L81

pkg/webhook/v1alpha1/placemanifestwork_validating.go
```diff
@@ -55,6 +55,12 @@ func (r *PlaceManifestWorkWebhook) validateRequest(newPlaceWork *workv1alpha1.Pl
...
+		return apierrors.NewBadRequest("Placement name should not be empty")
+	}
 
+	if len(placeWork.Spec.ManifestWorkTemplate.Workload.Manifests) == 0 {
+		return apierrors.NewBadRequest("ManifestWorkTemplate workload should not be empty")
+	}
```
qiujian16: you may also want to validate the payload in manifests by reusing the same func for manifestwork validation

pkg/webhook/v1alpha1/placemanifestwork_validating.go
serngawy: do you mean remove this validation and do it in the api repo ?

qiujian16: no I mean you could have a marker like this https://github.com/open-cluster-management-io/api/blob/main/addon/v1alpha1/types_clustermanagementaddon.go#L118-L120, that is sufficient to validate if name is empty.

serngawy: okay, just confirm wt to do

serngawy: done

pkg/webhook/v1/manifestwork_validating.go
```diff
@@ -76,16 +76,24 @@ func (r *ManifestWorkWebhook) validateRequest(newWork, oldWork *workv1.ManifestW
 }
 
 func validateManifests(work *workv1.ManifestWork) error {
+	return ValidateManifestWorkSpec(work.Spec)
+}
+
+func ValidateManifestWorkSpec(manifestWorkSpec workv1.ManifestWorkSpec) error {
```
qiujian16: nit the name of ValidateManifestWorkSpec is a bit confusing, we actually validate manifest in the spec.

qiujian16 (commented):
just a nit comment on naming,

/approve

pkg/webhook/v1/manifestwork_validating.go
serngawy: I name it after the given param manifestWorkSpec and there is a validateManifest func [Line 106](https://github.com/open-cluster-management-io/work/pull/182/files#diff-c05eb54e7b9381bee7827595cf0c481f6c38395bdafaccfb692136bfddba67aaR106)

qiujian16: I would remove validateManifests in line 78 and name this func as ValidateManifests. Will it be OK?

serngawy: okay, done

Issue #185, opened 2023-02-21
Manifest size limit


fgiloux:
I am wondering where the 50k limit comes from:
https://github.com/open-cluster-management-io/work/blob/main/pkg/webhook/manifestwork_webhook.go#L26-L27

[etcd default size limit](https://etcd.io/docs/v3.5/dev-guide/limit/) for request is 1.5MB
[kubernetes limit for configMaps](https://kubernetes.io/docs/concepts/configuration/configmap/#motivation) or secrets is 1MB

It means that ManifestWork cannot be used for large CRDs or ConfigMaps. This results in errors like:
~~~
E0221 09:04:14.759235       1 base_controller.go:159] "addon-deploy-controller" controller failed to sync "cluster1/olm-addon", err: [admission webhook "manifestworkvalidators.admission.work.open-cluster-management.io" denied the request: manifests should not be empty, admission webhook "manifestworkvalidators.admission.work.open-cluster-management.io" denied the request: the size of manifests is 269941 bytes which exceeds the 50k limit, admission webhook "manifestworkvalidators.admission.work.open-cluster-management.io" denied the request: the size of manifests is 99458 bytes which exceeds the 50k limit]
~~~

Also I am not sure of the value of this check. If the ManifestWork resource is too large, an error will get returned. What added value does the error returned by the validation webhook provide?

qiujian16:
Currently 50k is just a guess. We need the validation because in addition to manifests, we also record applied/available status of each manifest and optionally some status field of the applied resource if the user configure it. If a user create a manifestwork with a very large manifets, it is possible, that creating it succeeds, but agent cannot update its status.

We had an issue tracked before to allow configuring the size limit on manifestworks. FYI @zhiweiyin318

qiujian16:
https://github.com/open-cluster-management-io/addon-framework/issues/116 one of the fix we did is if you have several manifests and the total size is larger than 50k, it will be put into multiple manifestworks by addon-framework library using a bin-packing approach.

fgiloux:
[slack thread](https://kubernetes.slack.com/archives/C01GE7YSUUF/p1677510984910089)

qiujian16:
[...]

PR #186, opened 2023-02-28
the manifest size limit can be configured

zhiweiyin318:
issue: https://github.com/open-cluster-management-io/work/issues/185


qiujian16 (commented):
/approve
/lgtm

PR #191, opened 2023-03-03
Add e2e/integration test case for manifestworkreplicaset

qiujian16:


pkg/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_status_test.go
```diff
@@ -92,14 +92,14 @@ func TestStatusReconcileAsProcessing(t *testing.T) {
 
 	for id, cls := range clusters {
 		mw, _ := CreateManifestWork(mwrSetTest, cls)
-		cond := getCondition(string(workv1.ManifestApplied), "", "", metav1.ConditionTrue)
```
serngawy, line 95: I understand both workv1.ManifestApplied and workv1.WorkApplied have the same value but based on the API definition workv1.ManifestApplied is the condition type that we need to match for,  are we doing this to avoid string casting ?

qiujian16: workv1.WorkApplied is the condition for work.status.conditions, ManifestApplied is the condition per manifests.

serngawy (approved):

PR #195, opened 2023-03-28
fix spelling in err message

maleck13:
I was testing out the status options with manifestwork and noticed a small typo in the err message


qiujian16:
good catch, Thanks! Would you please signoff the commit?

```
git commit --amend -s
git push -f
```

maleck13:
That should be signed now. Nope did something wrong. Let me try that again. Actually I think it is correct now

PR #198, opened 2023-04-13
Fix available controller to reduce conflict

qiujian16:
1. requeue all works when resynce instead of handle them together in sync
2. start update condition only when finalizer and applied condition is set


skeeey:
refer to https://github.com/open-cluster-management-io/OCM/issues/130

PR #200, opened 2023-04-23
increase controller workers for available and cleanup

skeeey:


pkg/spoke/spokeagent.go
```diff
@@ -26,6 +26,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
+const (
+	// TODO expose a flag to set the worker for each controller
+	appliedManifestWorkFinalizeControllerWorkers = 10
```
qiujian16, line 36: can we add a comment on why it needs to be 10?

skeeey: added, PTAL

Issue #196, opened 2023-04-06
feedback rules does not work if there is multiple match


qiujian16:
An example is in Gateway API, we want to get all the addresses form an array:

```
addresses:
- type: IPAddress
  value: 172.32.201.0
- type: IPAddress
  value: 172.32.202.0
```

Using json path in feedback rule does not work, since today's implementation only return the first matched item.
```
feedbackRules:
- type: JSONPaths
  jsonPaths:
  - name: addresses
    path: 'status.addresses[?(@.type=="IPAddress")].value'
```

Possible way we could solve is this we could let jsonPaths to return a marshaled json string if the returned type is not a know scalar type. But to avoid overusing the json string which could be very large, we also need to set a reasonable size limit.

maleck13:
Yes. I was thinking something similar. Any thoughts on what would be a reasonable size limit? Is it something we would allow to modified from the hub?

qiujian16:
I would think a hard limit on this. Maybe start with 1K, and return a false condition if the returned value is larger than the limit.

deads2k:
I think CEL is a good fit for building a single string.  Another possible consideration is whether we want to allow retrieving more an one item?  The choices aren't necessarily mutually exclusive.

qiujian16:
< Another possible consideration is whether we want to allow retrieving more an one item

@deads2k  Yes I think it is desired. How do you think the idea of returning a json raw string with predefined limited size in this case?

maleck13:
So the specific use case (just for reference) is that a multi-cluster gateway controller in the hub is monitoring these gateway instances. These instances in the manifest work are based of a gateway definition created and placed in the hub. The controller needs to gather their addresses so that it can setup needed DNS records to bring traffic to the gateway(s).
As it is a controller, I had thought raw json could be a good option as that could be relatively simple to turn into an existing structure. I have heard of CEL but not used it in anger so not sure if CEL would be an advantage in this use case?

david-martin:
A couple of example implementations linked from the [multicluster Work API page](https://multicluster.sigs.k8s.io/concepts/work-api/) 

https://github.com/karmada-io/api/blob/549df6698be116aa9785242e9bc288bc94c2d59f/work/v1alpha1/work_types.go#L58-L91
provides an array of ManifestStatus (which is a runtime.RawExtension) on the Work status

[...]

PR #202, opened 2023-05-22
Allow return rawjson when featuregate is enabled

qiujian16:
relates to https://github.com/open-cluster-management-io/work/issues/196

the feature is enabled when setting RawFeedbackJsonString=true in work agent's feature gate


pkg/spoke/statusfeedback/reader.go
```diff
@@ -121,6 +127,24 @@ func getValueByJsonPath(name, path string, obj *unstructured.Unstructured) (*wor
...
 		}, nil
+	default:
+		if features.DefaultSpokeMutableFeatureGate.Enabled(ocmfeature.RawFeedbackJsonString) {
+			jsonRaw, err := json.Marshal(&t)
+			if err != nil {
+				return nil, fmt.Errorf("failed to parse the resource to json string: %v", err)
+			}
+			if len(jsonRaw) > maxJsonRawLength {
+				return nil, fmt.Errorf("the length of returned json raw string is larger than the maximum length %d", maxJsonRawLength)
+			}
```
elgnay: It will be clearer if the error message includes the name of the JSON path.

maleck13:
@qiujian16 great to see this change landing so quickly. Once it merges will give it a try. Is the work agent feature gate a start up flag?

qiujian16:
> @qiujian16 great to see this change landing so quickly. Once it merges will give it a try. Is the work agent feature gate a start up flag?

yes, in the latest version of clusteradm, you can run `clustadm join ... --feature-gates=RawFeedbackJsonString=true` to enable it

